PR #30, opened 2026-07-23
feat(core): add stderr-style structured-log stream sink

that1guy15:
## Summary

**W02 implementation PR-4 of 5** toward G02. Adds **`StreamLogSink`** — a `StructuredLogSink` that
writes exactly the CanonicalJSONV1 event-line bytes (from PR-2's `structured_log_event_line`) to an
**injected binary stream**, per A02 §4.15 ("stderr uses the exact event-line bytes without file header,
trailer, or sequence").

- **Fail-closed:** the `local_log` egress guard runs inside the projection *before* any byte is
  written, so a denied/widened matrix emits **nothing**.
- **Leak-safe:** a hostile stream failure wraps to a fixed `MH_LOG_SINK_WRITE` with no detail and
  `__cause__` cleared; non-writable stream → `MH_LOG_SINK_STREAM`; non-event → `MH_LOG_SINK_EVENT`.
- Structurally satisfies the `StructuredLogSink` Protocol, so a `StructuredLogger` can deliver to it.

## Proof-first tests
- `tests/unit/test_log_sink.py` — exact-one-line bytes, sequential append, input validation, hostile
  stream-failure wrapping, and a `StructuredLogger` integration test asserting **stdout/stderr stay
  empty** (capsys).
- `tests/security/test_log_sink_boundary.py` — planted-secret (fingerprint + error message) never
  reaches the emitted bytes; **fail-closed emits nothing** when egress denies.

## Evidence
- Independent report-only `milhouse-gate-review`: **pass**, no actionable findings.
- `make quality` ✅ · `make test` (3712 passed) ✅ · `make test-coverage` (line 98.23%, **branch
  97.18%**) ✅ · docs/skill/secret ✅ · `git diff --check` clean.

## Scope fences
No file I/O / rotation / persistence (**W03**); no CLI/stderr binding or flush/buffer policy (**W06**) —
writes only to an injected stream. No gate marked passed. Next: PR-5 (G02 evidence packet +
report-only review; owner accepts G02).

### DIFF
--- a/src/milhouse/core/log_wire.py
+++ b/src/milhouse/core/log_wire.py
@@ -2,11 +2,14 @@
 
 This is the W02-owned event-line wire: it maps a constructor-controlled
 ``StructuredLogEventV1`` to bounded CanonicalJSONV1 UTF-8 JSONL bytes with one trailing
-line feed. It performs no file or sink I/O (that is W03) and carries no arbitrary-text or
-exception-detail field.
+line feed, plus the injected-stream ``StreamLogSink`` that emits those exact bytes. It performs no
+file I/O, rotation, or persistence (that is W03) and carries no arbitrary-text or exception-detail
+field.
 """
 
 from __future__ import annotations
+
+from typing import Protocol
 
 from milhouse.core.canonical import canonical_json_bytes
 from milhouse.core.logging import LoggingError, StructuredLogEventV1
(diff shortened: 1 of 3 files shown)


that1guy15:
Codex independent review of exact head `8ab622006a21bab3c9cc6a97c1e561c04336c215`

Verdict: **changes required; do not merge yet.** Hosted CI is green, the focused sink suite passes, and DCO passes, but two locked G02 contracts are not satisfied.

### [P1] Normalize every hostile stream failure

`src/milhouse/core/log_wire.py:95,103-110`

- Capability lookup evaluates a potentially hostile `write` property outside error normalization.
- A stream-raised `LoggingError` is passed through verbatim and can carry arbitrary private detail.
- `raise ... from None` suppresses display of an ordinary source exception but retains it in `__context__`, so the detail is still reachable through the exception object.

This contradicts plan section 4.15: log failures must contain only fixed safe `MH_LOG_*` metadata, with no sensitive detail in exceptions or tracebacks.

Required correction: normalize constructor capability lookup and every stream-originated `Exception`, including `LoggingError`. Raise the fixed error only after leaving the exception handler so both `__cause__` and `__context__` are empty. Keep projection and egress errors outside the stream-write handler.

Required regressions: canary-bearing `OSError`, stream-raised `LoggingError`, and hostile `write` descriptor/property. Assert the canary is absent from `str`, `repr`, formatted traceback, and the recursive cause/context graph; assert both cause and context are absent.

### [P1] Do not acknowledge partial writes as complete lines

`src/milhouse/core/log_wire.py:78-80,103-104`

The protocol permits an arbitrary return value and the implementation ignores it. A stream may legally return zero or fewer than `len(line)` bytes without raising, after which `StreamLogSink.write` reports success and `StructuredLogger` can acknowledge a truncated or missing JSONL line. This contradicts section 4.15, where a successful emit means one complete event line was appended.

Required correction: implement bounded write-all behavior or fail closed on every non-full or invalid return. If partial-write retries are supported, serialize the entire operation with sink-owned locking so concurrent writers cannot interleave one event line.

Required regressions: zero, negative, oversized, `None`/`bool`, short-then-complete, repeated-short, and short-then-fail streams. Success must prove the exact canonical LF-terminated bytes were written once; no invalid or incomplete write may report success.

No CLI/stderr binding, filesystem persistence, provenance, private-data, or unrelated scope drift was found in this diff. Existing baseline document findings were not counted against PR #30.

that1guy15:
Both P1s addressed in `acabb18` (independent report-only re-review: pass; all gates green — branch coverage 97.19%).

### [P1] Normalize every hostile stream failure — fixed
- **Constructor capability lookup** is now inside error normalization: `getattr(stream, "write", None)` runs in a `try/except`, and `MH_LOG_SINK_STREAM` is raised **after** the handler unwinds — a hostile `write` descriptor/property that raises cannot leak.
- **Stream-raised `LoggingError` is no longer passed through.** The old `except LoggingError: raise` is gone; every stream-originated `Exception` (incl. `LoggingError`) is normalized to the fixed `MH_LOG_SINK_WRITE`.
- **Raised outside the handler.** `_write_all` catches internally and returns a bool; `write()` raises the fixed error only after the handler unwinds, so **both `__cause__` and `__context__` are empty**.
- Projection and egress errors stay **outside** the stream-write handler and keep their fixed codes (`MH_LOG_WIRE_EGRESS` still propagates).
- **Regressions:** canary-bearing `OSError`, stream-raised `LoggingError`, and hostile `write` descriptor — each asserts the canary is absent from `str`/`repr`/`args`/formatted-traceback **and** `__cause__ is None` **and** `__context__ is None`.

### [P1] Do not acknowledge partial writes as complete — fixed
- `StreamLogSink` now performs a **bounded write-all under a sink-owned `threading.Lock()`**, consuming the return value each iteration and failing closed on any zero, negative, oversized, or non-integer (`None`/`bool`) return. A partial or missing JSONL line is never reported as complete; the lock prevents concurrent writers from interleaving one event line.
- **Regressions:** `zero`, `negative`, `oversized`, `None`, `bool`, `short-then-complete`, `repeated-short`, `short-then-fail`. Success proves the exact canonical LF-terminated bytes were written once (`bytes(stream.written) == line`).

No CLI/stderr binding, filesystem persistence, or scope drift was added; the only new import is `threading`. Ready for re-review.

PR #38, opened 2026-07-24
feat(state): W03 SQLite control-plane foundation (slice 1)

that1guy15:
## Summary

First W03 slice: the SQLite control-plane substrate (ADR 0004, plan §4.3/§4.4) that the durable spool, replay, and retention build on. New `milhouse.state` package.

## Delivers
- **`database.py`** — secure `open_control_database`: DB created `0600` with `O_EXCL`/`O_NOFOLLOW` in an already-secured private parent, opened WAL + `synchronous=FULL` + enforced foreign keys, validated non-symlink owner-only; a `BEGIN IMMEDIATE` transaction boundary.
- **`migrations.py`** — ordered, checksum-enforced, contiguous migrations applied with their ledger row in one transaction; tampered/non-prefix ledger = corruption, fails closed.
- **`leases.py`** — named holder-owned time-bounded SQLite leases (acquire/renew/release), injected UTC time, lexical RFC3339 expiry.
- **`barrier.py`** — cross-process readers-writer global commit barrier: exclusive side is `filelock.FileLock` (BSD `flock LOCK_EX`), shared writer side is `fcntl.flock LOCK_SH` on the same file (they conflict correctly because both use BSD flock).

## Safety
Every failure raises a fixed `MH_STATE_*` error **outside any handler** — no SQLite/OS/caller detail in cause, context, args, or traceback. Fail-closed boundary tests inject filesystem/SQLite failures and **caught + fixed two real `__context__` leaks** (barrier exclusive-acquire and DB path-examine). 40 tests: golden behavior, injected-failure matrix, and a cross-process barrier exclusion/recovery proof.

## Notes
`filelock` was already a declared dependency (no dep change). `make quality` + `test` pass (3809 passed); global coverage floor holds. State modules are at ~94% branch; they'll join the enforced 95% `--critical` set in the W03 failure-injection slice, mirroring how `log_wire.py` joined critical at its evidence stage.

**Owner-confirmed W03 plan:** 7 fine-grained slices; this is slice 1 of 7 toward G03. G03 remains not-accepted.

### DIFF
--- a/src/milhouse/state/__init__.py
+++ b/src/milhouse/state/__init__.py
@@ -5,19 +5,29 @@ from __future__ import annotations
 from milhouse.state.barrier import GlobalCommitBarrier
 from milhouse.state.database import ControlDatabase, open_control_database
 from milhouse.state.errors import StateError
-from milhouse.state.leases import Lease, acquire_lease, release_lease, renew_lease
-from milhouse.state.migrations import Migration, apply_migrations, schema_version
+from milhouse.state.leases import (
+    Lease,
+    acquire_lease,
+    release_lease,
+    renew_lease,
+    require_current_lease,
+)
+from milhouse.state.migrations import Migration, migrate, schema_version
+from milhouse.state.schema import CONTROL_MIGRATIONS, initialize_control_state
 
 __all__ = [
+    "CONTROL_MIGRATIONS",
     "ControlDatabase",
     "GlobalCommitBarrier",
     "Lease",
     "Migration",
     "StateError",
     "acquire_lease",
-    "apply_migrations",
+    "initialize_control_state",
+    "migrate",
     "open_control_database",
     "release_lease",
     "renew_lease",
+    "require_current_lease",
     "schema_version",
 ]
--- a/src/milhouse/state/__init__.py
+++ b/src/milhouse/state/__init__.py
@@ -1,0 +1,23 @@
+"""W03 control-plane state: secure SQLite database, migrations, leases, and the commit barrier."""
+
+from __future__ import annotations
+
+from milhouse.state.barrier import GlobalCommitBarrier
+from milhouse.state.database import ControlDatabase, open_control_database
+from milhouse.state.errors import StateError
+from milhouse.state.leases import Lease, acquire_lease, release_lease, renew_lease
+from milhouse.state.migrations import Migration, apply_migrations, schema_version
+
+__all__ = [
+    "ControlDatabase",
+    "GlobalCommitBarrier",
+    "Lease",
+    "Migration",
+    "StateError",
+    "acquire_lease",
+    "apply_migrations",
+    "open_control_database",
+    "release_lease",
+    "renew_lease",
+    "schema_version",
+]
--- a/src/milhouse/state/barrier.py
+++ b/src/milhouse/state/barrier.py
@@ -1,0 +1,130 @@
+"""The global commit barrier: shared for durable writers, exclusive for maintenance (W03 slice 1).
+
+Plan section 4.3 and ADR 0004 require every durable writer to hold the *shared* side of one global
+commit barrier while backup, restore, migration, and declared maintenance take the *exclusive* side.
+That is a cross-process readers-writer lock: many writers coexist, and the exclusive side waits for
+all of them to release and blocks new ones.
+
+The exclusive side is a ``filelock.FileLock`` (BSD ``flock`` ``LOCK_EX``); the shared side is
+``fcntl.flock`` ``LOCK_SH`` on the same lock file. ``filelock`` exposes only exclusive locks, so it
+cannot express the shared writer side alone, but both use BSD ``flock`` so they conflict correctly
+across processes. The barrier is advisory, POSIX-only, and fails closed without ``flock``.
+Contention on a non-blocking acquisition raises a fixed ``MH_STATE_BARRIER_BUSY``.
+"""
+
+from __future__ import annotations
+
+import fcntl
+import os
+from collections.abc import Iterator
+from contextlib import contextmanager
+from pathlib import Path
+from typing import NoReturn
+
+from filelock import FileLock, Timeout
+
+from milhouse.config.filesystem import lexical_absolute_path
+from milhouse.state.errors import StateError
+
+_LOCK_FILE_MODE = 0o600
+_NOFOLLOW = getattr(os, "O_NOFOLLOW", 0)
+_CLOEXEC = getattr(os, "O_CLOEXEC", 0)
+
+
+def _fail(code: str, message: str) -> NoReturn:
+    raise StateError(code, message)
+
+
+class GlobalCommitBarrier:
+    """A cross-process readers-writer commit barrier keyed by one advisory lock file."""
+
+    __slots__ = ("_path",)
+
+    def __init__(self, lock_path: str | Path) -> None:
+        if _NOFOLLOW == 0:  # pragma: no cover - Milhouse supports only no-follow POSIX hosts
+            _fail("MH_STATE_UNSUPPORTED", "the commit barrier requires no-follow file support")
+        self._path = lexical_absolute_path(lock_path)
+
+    @property
+    def path(self) -> Path:
+        return self._path
+
+    def _ensure_lock_file(self) -> None:
+        flags = os.O_CREAT | os.O_EXCL | os.O_WRONLY | _NOFOLLOW | _CLOEXEC
+        descriptor: int | None = None
+        failed = False
+        try:
+            descriptor = os.open(self._path, flags, _LOCK_FILE_MODE)
+        except FileExistsError:
+            failed = False
+        except OSError:
+            failed = True
+        finally:
+            if descriptor is not None:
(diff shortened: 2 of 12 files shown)


that1guy15:
## Independent W03 Slice 1 review

**Candidate:** `faac48dad2a14b80bc6157a24627b2b4e5c02c0d`  
**Verdict:** **fail** — the overall slice boundary is right, but two foundation defects can corrupt state or bypass the migration contract, and three concurrency/error-contract issues need resolution before this becomes the substrate for later W03 work.

The hosted Required CI matrix is green. I also independently ran the 40 new state tests: **40 passed**. The findings below come from additional adversarial probes against this exact head.

### [P1] The barrier can truncate a foreign hard-linked inode

**Action:** Change  
**Gate/contract:** ADR 0004 global commit barrier; W03 restrictive permissions and safe durable state.  
**Evidence:** `GlobalCommitBarrier._ensure_lock_file()` treats any `FileExistsError` as acceptable without validating the existing inode. `exclusive()` then gives that pathname to `FileLock`, which opens and truncates the inode after locking it. I created a `0600` file containing `important-control-state`, hard-linked it as `commit.lock`, and entered `exclusive(blocking=False)`; the original file became zero bytes. A pre-existing `0666` lock file is also accepted by `shared()` and remains `0666`.  
**Impact:** A stale, accidental, or hostile same-user hard link can turn maintenance acquisition into control-state destruction. The API also does not establish the permission/inode invariant later W03 code will rely on.  
**Exact correction:** Do not reopen an unchecked pathname through `FileLock`. Securely open the parent and leaf descriptor-relative/no-follow, validate owner, regular-file type, exact `0600`, single link, safe ACL, and stable parent/file identity, then use that same descriptor for both `LOCK_SH` and `LOCK_EX`. Raw `fcntl.flock` for both sides is the simplest way to avoid library truncation/reopen behavior. Fail closed for unsupported flock; do not fall back to a soft/existence lock.  
**Required regression evidence:** hard link to a populated file, symlink, FIFO/non-regular file, loose mode, unsafe ACL where supported, inode replacement, and normal cross-process SH/EX exclusion. Prove the foreign file bytes never change.

### [P1] Lease schema bypasses both the migration ledger and the exclusive barrier

**Action:** Change  
**Gate/contract:** Plan §4.3 requires migrations to take the exclusive global barrier; §4.4 requires transactional, checksum-recorded schema migrations.  
**Evidence:** `leases._ensure_table()` creates `_leases` ad hoc on every lease operation. On a pristine database, calling `acquire_lease()` and then `schema_version()` returns version `0` while both `_leases` and `_schema_migrations` exist. `apply_migrations()` is also exported without acquiring or requiring the exclusive barrier, and `schema_version()` mutates the database by creating the ledger table.  
**Impact:** The physical schema can diverge from the recorded schema version/checksums; later packaged migrations can collide with unversioned tables, and concurrent startup can perform DDL outside the maintenance fence. This defeats the migration foundation before spool/ledger tables are added.  
**Exact correction:** Define the initial control schema, including `_leases`, in an immutable package-owned migration sequence. Provide one startup migration entry point that owns the exclusive barrier for the whole inspect/apply operation. Lease operations must assume the migrated schema and perform no DDL. Make schema/status inspection read-only (absent ledger may report version 0 without creating it).  
**Required regression evidence:** pristine DB → explicit migration → exact version/checksum/tables; lease and status calls make no schema changes; altered/gapped migrations fail; two processes attempting startup migration serialize through the exclusive barrier; a failed migration leaves neither schema objects nor ledger row.

### [P2] Public state operations still leak raw backend and time exceptions

**Action:** Change  
**Gate/contract:** PR safety claim and `StateError` contract: state-owned failures are fixed `MH_STATE_*` errors with no backend detail/cause/context.  
**Evidence:** A malformed existing `_leases` table makes `acquire_lease()` raise raw `sqlite3.OperationalError('no such column: holder')`, not `StateError`. Passing a valid typed but overflowing instant/TTL raises raw `OverflowError('date value out of range')`. Similar unnormalized paths exist around schema-table DDL/load, lease queries/DML, and schema-version reads.  
**Impact:** Corrupt state and boundary inputs can escape the stable error surface and later reach CLI/log/diagnostic rendering with implementation detail. Recovery code cannot reliably branch on `MH_STATE_*` codes.  
**Exact correction:** Normalize every state-owned SQLite/filesystem/time failure at each public boundary, raising the fixed error only after leaving the handler. Preserve caller exceptions from a transaction body only where that behavior is intentional and documented. Bound TTL before arithmetic.  
**Required regression evidence:** malformed/missing tables and columns, query/DDL/DML failures for every public state operation, invalid/overflowing time inputs, and injected begin/commit/rollback/close failures; assert exact code and safe args plus `__cause__ is None` and `__context__ is None`.

### [P2] A queued exclusive maintenance operation does not stop new shared writers

**Action:** Change  
**Gate/contract:** ADR 0004 and plan §10.3 require backup/restore/migration to acquire the exclusive barrier, drain active durable writers, and briefly hold a stable snapshot boundary.  
**Evidence:** With one process holding `shared()`, I queued a second process in blocking `exclusive()`. While that exclusive waiter was blocked, a third `shared(blocking=False)` acquisition succeeded. Plain flock supplies exclusion once acquired but no admission gate for a waiting maintainer.  
**Impact:** Continuous scheduler/receiver/CLI writer arrivals can starve backup, restore, migration, or purge rather than allowing the bounded maintenance drain expected by later packages.  
**Exact correction:** Add a cross-process writer-preference admission/turnstile protocol (or another explicitly bounded maintenance-drain mechanism) so once exclusive maintenance is queued, later shared entrants wait behind it while existing shared holders drain.  
**Required regression evidence:** cross-process test with shared holder A, queued exclusive B, then shared C; C must not enter before B acquires and releases. Also prove multiple shared holders still coexist when no exclusive operation is pending and crash release restores progress.

### [P2] The lease has no fencing generation for stale-holder rejection

**Action:** Change  
**Gate/contract:** W03 multi-process coordination foundation; W15 later requires two schedulers never to overlap one job.  
**Evidence:** `Lease` contains only name, holder, acquisition, and expiry. On expiry, a new holder overwrites the row; the prior holder retains no generation/token that subsequent protected mutations can compare atomically. An old process can therefore resume work after takeover even though renewal would now fail.  
**Impact:** TTL expiry alone cannot provide the future no-overlap guarantee or prevent stale writers after process suspension/clock delay. Retrofitting fencing after spool/ledger consumers exist would require a wider schema/API change.  
**Exact correction:** Add a monotonic lease generation/fencing token that increments on ownership change and remains stable on renewal. Any lease-protected SQLite mutation must validate name/holder/token in the same transaction. Document the later scheduler renewal/cancellation rule for work with external side effects.  
**Required regression evidence:** holder A expires, holder B takes over with a greater token, and A cannot renew, release B, or perform a token-guarded mutation; same-holder renewal retains the token; concurrent contenders produce one winner.

### What is sound

The slice is scoped in the right place: SQLite open/transactions, immutable migrations, leases, and the global barrier are exactly the primitives W03 needs before spool code. WAL + FULL synchronous mode, injected time, small modules, DCO, and the cross-process test direction are all good. I would keep this slice small and repair these primitives here rather than letting later spool/recovery slices build on ambiguous filesystem, migration, and lease semantics.

G03 remains pending; the deferred 95% critical-module coverage floor is not itself a blocker for this first slice, but the corrected boundary branches should be included when these modules enter the critical set.

that1guy15:
## Review addressed (commit `d7c4a48`)

All five findings are fixed with the required regression evidence; no source path outside `milhouse.state` changed.

**[P1] Barrier hard-link truncation** — the barrier no longer uses `filelock` (which reopened + truncated the pathname). Both sides now use `fcntl.flock` on a securely opened, **read-only, no-follow** descriptor validated as owner-only/single-link/regular (via the W02 secure-open machinery); the inode is never reopened for writing. Regression: a hard link to a populated file, a symlink, a FIFO, and a loose-mode file all fail closed, and the hard-linked file's bytes are **proven byte-for-byte intact**.

**[P1] Lease/schema bypass** — the initial schema (incl. a fenced `_leases`) is now an immutable **package migration** (`CONTROL_MIGRATIONS`). New `migrate` owns the **exclusive barrier** for the whole inspect-and-apply; `initialize_control_state` is the startup entry; lease ops do **no DDL**; `schema_version` is **read-only** (absent ledger → 0, no create).

**[P2] Raw exception leaks** — every state-owned SQLite/time failure is normalized to a fixed `MH_STATE_*` raised outside the handler; a malformed `_leases` table and an overflowing deadline are covered, and the TTL is bounded before arithmetic.

**[P2] Writer preference** — a two-file **turnstile**: a queued exclusive holds the gate for its duration so new shared entrants block while existing writers drain. Proven by a cross-process test (shared A, queued exclusive B, refused shared C).

**[P2] Lease fencing** — `_leases` carries a monotonic `fence` token (increments on ownership change, stable on renewal); `require_current_lease` validates name/holder/fence in the caller's transaction so a taken-over holder cannot commit.

`make quality`/`test` pass (3819 passed); global coverage floor holds. State modules ~91% branch; they join the enforced 95% `--critical` set in the W03 failure-injection slice.

PR #40, opened 2026-07-24
feat(spooling): W03 durable segment commit + ledger (slice 3a)

that1guy15:
## Summary

Reconcile a published segment with the SQLite control plane (ADR 0004). This is the **durable-commit half** of the plan's slice 3 (commit + recovery); reconciliation/recovery follows as slice 3b. Slice **3a of 7**.

## Delivers
- Control **migration 2** adds the `_segments` ledger (batch id, day, scope/target, privacy, retention, count, ordered-frame digest, byte size, whole-file digest, committed-at, delivery status).
- **`commit_segment`** — holding the **shared** commit barrier, atomically publishes the segment file under `<spool_root>/pending/<day>/` (W02 secure write) then inserts its ledger row and commits SQLite. A crash/ledger-failure after publication leaves a durable **orphan** for reconciliation and is reported as fixed commit-uncertain `MH_SPOOL_COMMIT`; a duplicate batch fails closed at publication.
- **`read_segment`/`list_segments`** read-only queries; `build_segment_bytes` factored out so the commit can compute size + whole-file digest.

## Pre-review (adversarial, before pushing)
Ran a 4-lens review workflow; it found and this change fixes:
- **[P2]** a failing BEGIN/COMMIT raises `StateError` (not `sqlite3.Error`), so a genuinely commit-uncertain post-publication ledger failure escaped as `MH_STATE_TXN` instead of `MH_SPOOL_COMMIT` — now caught + relabeled, with a regression test.
- **[P3]** malformed-row `int()` coercion moved inside the fail-closed boundary (+ test); directory owner check uses the effective uid to match the W02 writer; directories required to be exactly `0o700`.

## Verification
`make quality`/`test` pass (**3859 passed**), global floor holds; DCO clean. Spooling modules join the enforced 95% critical set in the W03 failure-injection slice.

**Known minor gap (disclosed):** commit.py validates each dir's mode/owner/no-symlink but relies on the W02 writer for the segment file's immediate-parent ACL check; an extended ACL on `spool_root` itself isn't separately inspected (the `0o700` no-ACL day dir shields the `0o600` file). Can tighten in slice 3b.

### DIFF
--- a/src/milhouse/spooling/__init__.py
+++ b/src/milhouse/spooling/__init__.py
@@ -2,6 +2,11 @@
 
 from __future__ import annotations
 
+from milhouse.spooling.commit import (
+    DurableSpool,
+    ExporterDelivery,
+    SegmentRecord,
+)
 from milhouse.spooling.errors import SpoolError
 from milhouse.spooling.segment import (
     SegmentHeaderV1,
@@ -10,12 +15,21 @@ from milhouse.spooling.segment import (
     spool_frame_line,
     spool_segment_header_line,
 )
-from milhouse.spooling.writer import write_spool_segment
+from milhouse.spooling.writer import (
+    build_segment_bytes,
+    publish_segment_bytes,
+    write_spool_segment,
+)
 
 __all__ = [
+    "DurableSpool",
+    "ExporterDelivery",
     "SegmentHeaderV1",
+    "SegmentRecord",
     "SpoolError",
     "SpoolFrameV1",
+    "build_segment_bytes",
+    "publish_segment_bytes",
     "spool_content_sha256",
     "spool_frame_line",
     "spool_segment_header_line",
--- a/src/milhouse/spooling/commit.py
+++ b/src/milhouse/spooling/commit.py
@@ -1,0 +1,409 @@
+"""Durable spool commit and segment ledger, bound to one control plane (W03 slice 3a).
+
+A :class:`DurableSpool` binds one control database, one commit barrier, and one spool root that all
+live under the same canonical state root, so a caller cannot publish under a different barrier than
+maintenance holds. Per ADR 0004 and plan section 4.7 a commit is a deliberate, authorized,
+reconciled operation: it authorizes the ``local_spool`` and ``local_sqlite`` egress surfaces
+(restricted input is rejected before any mutation), derives the ``YYYY-MM-DD`` partition from the
+commit instant, then, while holding the shared commit barrier, creates the partition directory,
+atomically publishes the exact segment bytes (write, flush, fsync, no-replace rename, parent fsync),
+and inserts the segment ledger row plus one row per required exporter in a single SQLite
+transaction. A crash or ledger failure after publication leaves a durable orphan and is reported as
+commit-uncertain ``MH_SPOOL_COMMIT``, never as success. The ledger preserves every immutable header
+field so recovery never infers policy from newer configuration, and every read is semantically
+validated so a corrupt or foreign row fails closed with ``MH_SPOOL_LEDGER``. Every failure raises a
+fixed ``MH_SPOOL_*`` error outside the handler.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import os
+import re
+import sqlite3
+import stat
+from dataclasses import dataclass
+from datetime import datetime
+from pathlib import Path
+from typing import Any, NoReturn
+
+from milhouse.config.filesystem import lexical_absolute_path
+from milhouse.core.clock import TimeError, format_timestamp
+from milhouse.privacy import (
+    EgressDisposition,
+    EgressSurface,
+    PrivacyError,
+    require_egress,
+)
+from milhouse.spooling.errors import SpoolError
+from milhouse.spooling.segment import (
+    EXPORTER_ID_PATTERN,
+    FRAME_VERSION,
+    SCHEMA_VERSION,
+    SegmentHeaderV1,
+    SpoolFrameV1,
+)
+from milhouse.spooling.writer import build_segment_bytes, publish_segment_bytes
+from milhouse.state.barrier import GlobalCommitBarrier
+from milhouse.state.database import ControlDatabase
+from milhouse.state.errors import StateError
+
+_SPOOL = "spool"
+_BARRIER_NAME = "commit.lock"
+_PENDING = "pending"
+_DIR_MODE = 0o700
+_DELIVERY_PENDING = "pending"
+_DAY_LENGTH = 10
+_SCOPES = ("installation", "target")
+_ALLOWED_PRIVACY = ("public", "internal", "sensitive")
+_DELIVERY_STATES = ("pending", "delivered", "failed")
+_SHA256_HEX = re.compile(r"[0-9a-f]{64}", flags=re.ASCII)
+_STAMP = re.compile(
+    r"[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}\.[0-9]{3}Z", flags=re.ASCII
+)
+_SEGMENT_COLUMNS = (
+    "batch_id",
+    "day",
+    "schema_version",
+    "frame_version",
+    "config_generation",
+    "scope",
+    "target_id",
+    "privacy_class",
+    "retention_days",
+    "record_count",
+    "content_sha256",
+    "byte_size",
+    "file_sha256",
+    "committed_at",
+)
+
+
+def _fail(code: str, message: str) -> NoReturn:
+    raise SpoolError(code, message)
+
+
+def _current_uid() -> int:
+    # Match the W02 secure-file writer (filesystem.py), which validates against the effective uid.
+    geteuid = getattr(os, "geteuid", None)
+    if geteuid is None:  # pragma: no cover - Milhouse supports only POSIX hosts
+        _fail("MH_SPOOL_UNSUPPORTED", "the spool requires a POSIX ownership model")
+    return int(geteuid())
+
+
+@dataclass(frozen=True, slots=True)
+class ExporterDelivery:
+    """The delivery state of one required exporter for a committed segment."""
+
+    exporter_id: str
+    delivery_status: str
+
+
+@dataclass(frozen=True, slots=True)
+class SegmentRecord:
+    """A committed segment's immutable ledger row plus its per-exporter delivery rows."""
+
+    batch_id: str
+    day: str
+    schema_version: int
+    frame_version: int
+    config_generation: str
+    scope: str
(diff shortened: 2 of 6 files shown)


that1guy15:
## Independent W03 Slice 3a review

**Candidate:** `0bac9376c4742d6f65a1c4cea4b812810945ae95` on base `902b304ba9db284e98ed3b69145d8a458a267602`  
**Verdict:** **fail**

The publish → parent-fsync → ledger-commit ordering is the correct ADR 0004 shape, and hosted Required CI is green. I independently ran the integrated spool segment + commit tests: **39 passed**. Additional contract probes found three merge-blocking foundation defects and two bounded correctness defects.

### [P1] The durable spool and SQLite writes bypass the common egress authorization

**Action:** Change  
**Gate/contract:** Plan §4.7 and `PRIVACY.md`: every later persistence surface must invoke `require_egress`; restricted input may never reach local spool or SQLite.  
**Evidence:** Neither `writer.py` nor `commit.py` invokes `require_egress(EgressSurface.LOCAL_SPOOL, ...)` or `LOCAL_SQLITE`. `SegmentHeaderV1` explicitly accepts `privacy_class="restricted"`. I constructed a valid empty restricted header (`record_count=0`, SHA-256 of empty frame bytes) and called `commit_segment`; it created the JSONL file and a ledger row whose privacy class was `restricted`.  
**Impact:** This is a direct violation of the locked privacy ceiling. Relying on canonical-record validation is insufficient because an empty segment has no record to reject, and the sink itself remains an authorization bypass.  
**Exact correction:** Authorize both local-spool and local-SQLite surfaces before creating directories, publishing bytes, or opening a ledger transaction. Require the returned disposition to permit a redacted record. Reject `restricted` in the segment header itself and normalize `PrivacyError` to a fixed safe spool error outside its handler.  
**Required regression evidence:** matrix coverage for public/internal/sensitive/restricted on both persistence surfaces; an empty restricted segment must leave no directory, file, or ledger row; planted authorization failure must be context/cause-clean.

### [P1] The ledger does not preserve the immutable policy needed for recovery and delivery

**Action:** Change  
**Gate/contract:** Plan §4.3 requires SQLite to record matching immutable header metadata and says orphan recovery must never infer delivery policy from newer configuration; §4.4 requires segment and exporter delivery ledgers; `spool verify` must prove header/ledger agreement.  
**Evidence:** Migration 2 and `SegmentRecord` omit `config_generation`, `required_exporters`, `schema`, and `frame_version`. They store only one aggregate `delivery_status` and create no per-exporter rows. The successful commit transaction therefore cannot reconstruct or compare the exact header policy, and cannot represent partial delivery when a segment requires more than one exporter.  
**Impact:** Later recovery/export code must reread the file or consult current configuration to discover policy, cannot fully verify ledger agreement, and would require a backfill/schema repair after this supposedly immutable migration has merged. That is exactly the drift ADR 0004 forbids.  
**Exact correction:** Amend migration 2 before merge so the segment ledger preserves every immutable header field needed for exact agreement. Add a normalized `_segment_exporters` ledger keyed by `(batch_id, exporter_id)` with a foreign key, or an equivalently exact design. Insert the segment row and all required-exporter rows in the same SQLite transaction after publication. Treat aggregate segment delivery state as derived from exporter rows rather than the only authority.  
**Required regression evidence:** zero, one, and multiple required exporters; exact header↔ledger equality; independent per-exporter status; altered configuration after commit cannot change recovery/delivery policy; transaction failure leaves no partial segment/exporter ledger state and the published file remains a reconcilable orphan.

### [P1] The API cannot prove it is using the one global barrier, and it mutates the spool before acquiring it

**Action:** Change  
**Gate/contract:** Plan §4.3: every durable writer takes the shared side of the global commit barrier and no code path may bypass it; exclusive maintenance must observe a stable namespace.  
**Evidence:** `commit_segment` accepts an arbitrary database, barrier, and spool root as three unrelated arguments. It performs `_pending_day_dir(...)`—including `pending/` and day-directory creation—before entering `with barrier.shared()`. A caller can also pass a barrier rooted at an unrelated lock path while maintenance holds the canonical barrier.  
**Impact:** Backup, restore, purge, or migration can hold the intended exclusive barrier while a writer still mutates the spool namespace—or can publish and commit entirely under a different barrier. Later snapshot and purge correctness cannot be proven from this API.  
**Exact correction:** Move every spool namespace mutation inside the shared-barrier scope. Bind the verified control database, canonical barrier, and resolved spool root into one immutable runtime/store object constructed from the same verified `RuntimePaths` generation; `commit_segment` should consume that binding rather than independently supplied paths/locks. At minimum, reject a barrier/database/spool combination that is not under the same canonical state root.  
**Required regression evidence:** while canonical exclusive maintenance is held, a commit creates no directory, file, or row; a mismatched barrier or database is rejected before mutation; normal concurrent commits under the bound store remain noninterleaved.

### [P2] Semantically corrupt ledger rows are returned as valid records

**Action:** Change  
**Gate/contract:** W03 corruption must fail closed and retention/delivery state must not be inferred from malformed control data.  
**Evidence:** The added regression covers only values that fail `int()`. I inserted a row with `day="not-a-day"`, target scope plus null target, `privacy_class="restricted"`, negative retention/count/byte size, one-character digests, a non-timestamp commit value, and `delivery_status="unknown"`. Every coercion succeeded and `read_segment` returned a normal `SegmentRecord`. The migration has no `CHECK` constraints and `_row_to_record` performs no semantic validation.  
**Impact:** Future retention, exporter, verification, and recovery code can act on corrupt state as if it were authoritative, including unsafe expiry or false delivery decisions.  
**Exact correction:** Define one exact ledger-row validator and use it for all reads. Add SQLite `CHECK` constraints/strict typing where they materially prevent invalid state, while retaining read-time validation for a foreign or previously corrupted database. Validate enum values, safe IDs, scope/target relation, real UTC day/timestamp, positive bounds, 64-hex digests, nonnegative sizes/counts, and exporter/delivery states.  
**Required regression evidence:** adversarial mutation of every column and cross-field invariant; each invalid row must produce fixed `MH_SPOOL_LEDGER` with clean cause/context, and invalid normal writes must be rejected transactionally.

### [P2] The partition day is caller-controlled and need not be a real or matching date

**Action:** Change  
**Gate/contract:** The `pending/YYYY-MM-DD` layout must be deterministic and recoverable; the ledger path coordinate must agree with the authoritative commit instant.  
**Evidence:** `day` is an independent argument checked only by a digit regex. The probe above successfully committed under `pending/2026-02-31/` with `committed_at=2026-07-24T12:00:00.000Z`; the invalid date was also stored in SQLite.  
**Impact:** Equivalent commits can land in different partitions, invalid calendar directories become durable, and retention/recovery scans cannot rely on the path/ledger coordinate.  
**Exact correction:** Remove the independent `day` argument and derive the UTC partition from the already canonicalized commit timestamp (or document and lock another authoritative instant before implementation). Validate it through actual date semantics, not a regex.  
**Required regression evidence:** UTC day derivation across midnight, leap day, timezone-normalized inputs, invalid timestamps, and exact path/ledger agreement.

### [P3] Hash and publication should use one byte snapshot

**Action:** Consider  
**Gate/contract:** The ledger `byte_size` and `file_sha256` must describe the exact bytes published.  
**Evidence:** `commit_segment` builds and hashes `content`, then calls `write_spool_segment(header, frames)`, which independently rebuilds the bytes before publication. The current frozen/deterministic models make the outputs normally equal, but the invariant is duplicated rather than structural and the accepted `frames` list is mutable.  
**Impact:** Future serialization changes or concurrent mutation can turn an exact byte-binding guarantee into a race or unexplained failure.  
**Exact correction:** Build/validate once and pass those exact bytes to the secure atomic publisher, or have the writer return the exact published-byte metadata used for the ledger.  
**Required regression evidence:** prove serialization occurs once and the file bytes, byte size, and digest are derived from the same immutable snapshot.

### Non-blocking disposition

The disclosed spool-root ACL gap does not currently demonstrate content exposure because the W02 publisher validates the immediate day directory and its ACL before writing. It should still be tightened in Slice 3b as stated and included in the G03 filesystem matrix.

The slice boundary remains good: durable publication and the matching control-plane transaction belong together. These corrections should land before recovery is implemented, because recovery can only be as truthful as the persisted header/ledger/barrier contracts it inherits.

that1guy15:
## Fail-review addressed — reworked to a bound `DurableSpool`

Force-pushed a rework that resolves every finding, plus a second adversarial pass. Summary of what changed and the regression evidence added:

**P1 — egress bypass.** `commit_segment` now calls `_authorize(privacy_class)` *before any mutation*, requiring `REDACTED_RECORD` on both `local_spool` **and** `local_sqlite`; `restricted` is rejected. Tests cover public/internal/sensitive commit and a direct `_authorize("restricted") → MH_SPOOL_EGRESS`. (`restricted` is also unconstructible at the segment header, so it can never reach commit.)

**P1 — ledger omitted immutable policy/exporters.** Migration 2 now stores every immutable header field (schema/frame version, config_generation, scope, target_id, privacy_class, retention_days, content_sha256, byte_size, file_sha256, committed_at) under CHECK constraints, plus a new `_segment_exporters` table with one row per required exporter. Recovery never infers policy from live config. Tests assert field preservation and 0/1/N exporter rows with per-exporter status.

**P1 — barrier binding + mutate-before-barrier.** `DurableSpool` binds one db + barrier + spool_root under one state root; *all* mutation (mkdir, publish, ledger insert) is strictly inside `barrier.shared()`. A spy-barrier test asserts the pending dir does not exist when the barrier is entered.

**P2 — semantic read validation.** Every read routes rows through `_validated_segment`, failing closed with `MH_SPOOL_LEDGER`. A parametrized test mutates each column (day, versions, digests, scope/target relation, privacy, bounds, committed_at↔day) and asserts rejection.

**P2 — day derived, not caller-supplied.** The `YYYY-MM-DD` partition is derived from the commit instant (UTC-normalized); a naive instant fails closed with `MH_SPOOL_COMMIT`. Test covers a UTC-day rollover across a non-UTC offset.

**P3 — single byte snapshot.** `build_segment_bytes(...)` is computed once; size, digest, and the published file all come from that one `bytes`. Test asserts `file_sha256 == sha256(published bytes)`.

Also carried over from the first pre-review: `StateError` (not `sqlite3.Error`) at a failing txn boundary is caught and relabeled commit-uncertain; both post-publication failure tests assert `__cause__`/`__context__` are empty.

**Second adversarial pass** (new): the store now pins the barrier to the canonical `commit.lock` beside the database — a same-directory lock on a *different file* shares no `flock` with maintenance and is now rejected — and `exporter_id` shape is re-validated on read so a tampered exporter row fails closed.

Gate: `make quality`, `test-coverage`, and `package` pass — 3205 tests; line 97.89%, branch 96.67%; `commit.py` at 97% branch.

PR #41, opened 2026-07-24
feat(spooling): secure segment reader (W03 slice 3b)

that1guy15:
## W03 slice 3b — secure segment reader

The fail-closed **inverse of the segment writer**, per ADR 0004 and plan §4.3. Reconciliation, replay, and `spool verify` all need to read a durable segment file back and validate it against the writer — and no such reader existed (`DurableSpool.read_segment` reads the SQLite ledger, not the file). This slice adds it.

### What it does
- **`parse_segment_bytes(content, *, installation_id=None)`** — splits a self-describing segment into its header line and ordered frame lines, rebuilds the typed `SegmentHeaderV1` and `SpoolFrameV1` records, and **re-encodes every line with the canonical writer to prove it is byte-for-byte canonical** (a line that merely parses but is not canonical is rejected). It then verifies frame count, contiguous sequence, header/frame agreement, and the ordered-frame `content_sha256`, and returns the whole-file digest as a typed `ParsedSegment`.
- **`read_segment_file(path, *, require_private_parent=False, installation_id=None)`** — a W02 secure, no-follow, size-bounded read that distinguishes a missing file (`MH_SPOOL_NOT_FOUND`) and a symlink/non-regular file (`MH_SPOOL_NOT_REGULAR`) from corruption.
- Records reconstruct through **pydantic JSON mode** (`model_validate_json`) because the strict envelope refuses to coerce the wire's RFC3339 string timestamps in Python mode.
- Every failure is a fixed `MH_SPOOL_*` error raised **outside any handler**, so no filesystem, JSON, or record detail reaches its cause, context, args, or traceback.

### Adversarial review
Ran ~25 hostile inputs through the parser. Confirmed the three core guarantees — clean error context, byte-for-byte canonicity, bounded/no-DoS — then surfaced three findings, **all addressed**:

| | Finding | Resolution |
|---|---|---|
| P2 | Canonical form ≠ provenance: a canonical record carrying **another** record's valid id was accepted | optional `installation_id` re-derives each record's deterministic identity via `verify_record_identity` and fails closed; docstring no longer overstates the guarantee. Forged-record exploit is a regression test |
| P3 | The frame bound was on the header's attacker-**declared** count | now bounds the **real** frame list before the per-frame reconstruction |
| P3 | `require_private_parent` defaults `False` while the writer forces `True` | documented the trust dependency on `read_segment_file` |

### Tests & gate
41 reader tests: full round trip, empty segment, every corruption mode (truncation, blank/oversized lines, non-canonical header/frame, unexpected fields, count/digest/sequence/scope/target/privacy/batch-id disagreement, unreconstructable record), safety bounds, secure-open failures, an injected read error, and the forged-record exploit.

- `reader.py` **98% branch**
- `make quality`, `test-coverage`, `package` all pass — **3246 tests**; line **97.88%**, branch **96.70%**
- DCO-clean, single signed commit

### Next
Slice 3c — reconciliation + recovery (orphan register, missing-file→unhealthy, corrupt→quarantine, stale-tmp, torn-frame recovery, `spool verify`) — builds on this reader.

### DIFF
--- a/src/milhouse/spooling/__init__.py
+++ b/src/milhouse/spooling/__init__.py
@@ -8,6 +8,12 @@ from milhouse.spooling.commit import (
     SegmentRecord,
 )
 from milhouse.spooling.errors import SpoolError
+from milhouse.spooling.reader import (
+    ParsedSegment,
+    parse_segment_bytes,
+    read_trusted_segment,
+    read_untrusted_segment,
+)
 from milhouse.spooling.segment import (
     SegmentHeaderV1,
     SpoolFrameV1,
@@ -24,12 +30,16 @@ from milhouse.spooling.writer import (
 __all__ = [
     "DurableSpool",
     "ExporterDelivery",
+    "ParsedSegment",
     "SegmentHeaderV1",
     "SegmentRecord",
     "SpoolError",
     "SpoolFrameV1",
     "build_segment_bytes",
+    "parse_segment_bytes",
     "publish_segment_bytes",
+    "read_trusted_segment",
+    "read_untrusted_segment",
     "spool_content_sha256",
     "spool_frame_line",
     "spool_segment_header_line",
--- a/src/milhouse/spooling/reader.py
+++ b/src/milhouse/spooling/reader.py
@@ -1,0 +1,414 @@
+"""Segment reader: the fail-closed inverse of the segment writer (W03 slice 3b).
+
+Reconciliation, replay, and ``spool verify`` need to read a durable segment file back and validate
+it, even when the file is a crash artifact, a foreign write, or deliberately tampered. This module
+draws a hard line between two trust levels so a caller cannot accidentally accept untrusted bytes as
+durable state:
+
+* ``parse_segment_bytes`` is the **untrusted structural parser**. It splits the self-describing
+  segment into a header line and ordered frame lines, rebuilds the typed ``SegmentHeaderV1`` and
+  ``SpoolFrameV1`` records, and re-encodes each line with the canonical writer to prove it is
+  byte-for-byte canonical, then verifies frame count, contiguous sequence, header/frame agreement,
+  and the ordered-frame ``content_sha256``. It proves canonical form and internal self-consistency —
+  not provenance — so its result must be used only for diagnostics, quarantine classification, or
+  salvage, never accepted, registered, replayed, or exported as trusted durable state.
+* ``read_untrusted_segment`` reads such a file (no-follow, size-bounded) and structurally parses it,
+  for the same diagnostics-only purposes.
+* ``read_trusted_segment`` is the **only** acceptance path. It opens the file no-follow under an
+  owned ``0700`` parent, requires the descriptor to be a regular, owner-only ``0600``, single-link,
+  ACL-safe file, reads it under the size bound, re-checks the descriptor snapshot after the read to
+  reject any mutation during selection, and then **mandatorily** re-derives every record's
+  deterministic identity against a required ``installation_id`` so a canonical-but-forged or foreign
+  record fails closed. There is no optional argument that can bypass provenance.
+
+Every failure is a fixed ``MH_SPOOL_*`` error raised outside any handler, so no filesystem, JSON,
+record, or caller-supplied identity detail reaches its cause, context, args, or traceback. The
+reader never mutates; quarantine, orphan registration, and torn-frame recovery are later slices that
+build on the fixed error codes this reader raises to classify a corrupt file.
+"""
+
+from __future__ import annotations
+
+import hashlib
+import json
+import os
+import re
+import stat
+from dataclasses import dataclass
+from pathlib import Path
+from typing import NoReturn
+
+from milhouse.config.filesystem import (
+    FileSnapshot,
+    SecureFileError,
+    SecureFileErrorKind,
+    open_regular_file_no_follow,
+)
+from milhouse.domain.records import (
+    RecordEnvelopeV1,
+    RecordError,
+    verify_record_identity,
+)
+from milhouse.spooling.errors import SpoolError
+from milhouse.spooling.segment import (
+    FRAME_VERSION,
+    MAX_FRAME_LINE_BYTES,
+    MAX_HEADER_LINE_BYTES,
+    SCHEMA_VERSION,
+    SegmentHeaderV1,
+    SpoolFrameV1,
+    spool_content_sha256,
+    spool_frame_line,
+    spool_segment_header_line,
+)
+
+_LF = b"\n"
+# Reader safety bounds. A segment can never exceed these regardless of what its header claims; the
+# file size is checked from the descriptor before any bytes are read, so an oversized file is
+# rejected fast.
+MAX_SEGMENT_FILE_BYTES = 64 * 1024 * 1024
+MAX_SEGMENT_FRAMES = 1_000_000
+_TRUSTED_FILE_MODE = 0o600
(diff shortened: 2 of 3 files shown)


that1guy15:
Independent gate review — **fail** at `eb82b631b353a32efa0cf655a7bb271eaf479f08`.

The canonical JSONL reconstruction, count/sequence/header agreement, digest verification, size bounds, and no-follow open are well covered. The focused suite passes locally (`41 passed`), and every Required CI check is green on the reviewed head. Three trust-boundary defects still need correction before merge.

[P2] The recovery-facing reader silently skips record provenance by default  
Action: Change  
Gate/contract: Plan §4.3 requires `spool verify` to validate IDs and makes the durable spool the replay/reconciliation authority; ADR 0004 requires conflict-safe recovery.  
Evidence: `read_segment_file(..., installation_id=None)` forwards the omission to `parse_segment_bytes`, which only calls `_verify_identity` when the value is present (`reader.py:227-253, 307-324`). The new regression explicitly proves that a canonical-but-forged record is accepted when the argument is omitted (`test_spool_reader.py:392-409`). The PR describes this function as the reader for reconciliation, replay, and `spool verify`, so “callers should pass it” leaves the primary trust decision optional.  
Impact: A future recovery/replay caller can omit one keyword and register or replay a canonical foreign/forged record as trusted durable state. Documentation does not make that path fail closed.  
Exact correction: Make installation identity mandatory for the trusted file-reader/recovery API and bind that API to the configured `DurableSpool`/runtime context. If a structural parser is still needed for quarantine diagnostics or salvage, separate/name it explicitly as untrusted and do not expose its result to acceptance, reconciliation, replay, or ledger registration without mandatory provenance verification.  
Required regression evidence: The trusted reader must reject the canonical forgery with no optional argument that can bypass the check; recovery/replay/verify call paths must demonstrate they cannot consume a structurally parsed segment before provenance succeeds.

[P2] The “secure” file reader accepts loose, hard-linked, and mutation-unstable segment files  
Action: Change  
Gate/contract: G03 requires files/directories to be `0600`/`0700`; the spool is an immutable durable authority; the project’s secure-file contract uses descriptor identity to prevent replacement/mutation races.  
Evidence: `require_private_parent` defaults to `False`; even when true, `_read_all_bytes` retains only the descriptor, discards `OpenedRegularFile.selection.snapshot`, and never validates file owner, exact `0600` mode, link count, or a post-read snapshot (`reader.py:262-304, 307-324`). Direct probes against the reviewed tree produced:
- `DEFAULT_0777_PARENT_0644_FILE=ACCEPTED`
- `MODE_0644_PRIVATE_PARENT=ACCEPTED`
- `HARDLINKED_PRIVATE_FILE=ACCEPTED`
The last file had `st_nlink == 2`.  
Impact: Reconciliation can treat a file outside the locked spool namespace, a permissions-violating file, a second hard-link alias, or bytes changed during selection/read as an immutable committed segment. That undermines the recovery authority this slice is intended to establish.  
Exact correction: Provide a trusted reader bound to the canonical spool root and require the owned `0700` directory chain unconditionally. Validate the opened descriptor as owned, exact `0600`, regular, single-link, ACL-safe, and compare the captured `FileSnapshot` with a final `fstat` after the read. Fail with a fixed code if metadata changes. Keep any permissive raw-file inspection helper isolated from trusted recovery.  
Required regression evidence: Reject a `0644` segment, world-writable parent, hard link, ACL, wrong owner where the host permits the fixture, and an in-place append/truncate/same-size rewrite during read; prove the trusted path accepts the writer-produced `0600` segment.

[P2] Invalid installation identity escapes the fixed-error boundary  
Action: Change  
Gate/contract: The PR claims every failure is a fixed `MH_SPOOL_*` error with no record detail in cause, context, args, or traceback.  
Evidence: `_verify_identity` catches only `RecordError` (`reader.py:183-198`). Running `parse_segment_bytes(valid_content, installation_id="bad")` raises a raw Pydantic `ValidationError` beginning `1 validation error for MilhouseIdentityV1`, including rejected-input diagnostics, rather than `SpoolError`.  
Impact: The public API violates its advertised stable error contract and can disclose caller-supplied identity text through an exception path.  
Exact correction: Validate/normalize the installation identity at the reader boundary and contain all expected identity-validation failures inside the handler, then raise one fixed safe `MH_SPOOL_*` error outside it with empty cause/context. Prefer accepting the already-validated installation identity type in the trusted bound reader.  
Required regression evidence: Invalid and adversarial installation-ID inputs must yield only the fixed safe error; assert code, args, cause, context, string, repr, and traceback contain no planted canary.

Missing or externally pending gate evidence: none needed to assess this slice. G03 remains pending by design; this PR must not be treated as G03 acceptance.

Verdict: **fail**.

that1guy15:
## Gate-review defects addressed — force-pushed at `1ca9de5`

Thanks — all three P2 trust-boundary defects are fixed by reworking the reader around an explicit trust taxonomy. Summary:

**[P2] Provenance was optional by default.** The reader is now split so acceptance cannot skip provenance:
- `parse_segment_bytes(content)` — untrusted **structural** parser (no identity param), documented for diagnostics/quarantine/salvage only.
- `read_untrusted_segment(path)` — untrusted file read (no-follow + size-bound) for the same purposes.
- `read_trusted_segment(path, *, installation_id)` — the **only** acceptance path; `installation_id` is keyword-only with no default and `_verify_identity` runs unconditionally before every return. A call without it is a `TypeError` (regression test `test_trusted_read_cannot_bypass_provenance`). No path returns a trusted-looking `ParsedSegment` without provenance.

**[P2] The "secure" reader accepted loose/hard-linked/mutation-unstable files.** `read_trusted_segment` now:
- opens no-follow under an owned `0700` parent unconditionally (`require_private_parent=True`);
- validates the descriptor as regular, `st_uid == euid`, exact `0600`, `st_nlink == 1`, ACL-safe (parent/ACL via the open layer; leaf mode/owner/nlink via `_is_trusted_regular_file`) **before reading any bytes**;
- captures the open-time `FileSnapshot` and re-`fstat`s after the read, failing `MH_SPOOL_CHANGED` on any change. A same-size in-place rewrite is caught via `changed_ns` (ctime), which `utime` can't suppress.
- Regressions: `0644` file, world-writable parent, hard link (`nlink==2`), and mid-read same-size rewrite all rejected; the writer-produced `0600` single-link file is accepted.

**[P2] Invalid installation id escaped as a raw pydantic error.** `_require_installation_id` validates the id against the `InstallationIdV1` shape at the boundary (first statement of `read_trusted_segment`), failing `MH_SPOOL_IDENTITY` with a static message containing none of the input. A well-formed-but-foreign id still reaches `verify_record_identity` → `MH_SPOOL_RECORD`. Regression `test_trusted_read_rejects_a_malformed_installation_id_without_leaking_it` asserts the canary appears in no part of the error surface (code/args/str/repr/cause/context), and `test_reader_installation_pattern_matches_the_domain` locks the reader regex to the domain's `InstallationIdV1` so this can't silently regress.

A second adversarial re-review confirmed all three fixes with empirical probes (same-size rewrite + mtime spoof, hard-link ordering, ACL, foreign id) and found no new P1/P2.

**Scope note:** binding the trusted reader onto the configured `DurableSpool`/runtime context (so callers receive the installation identity from config rather than passing it) lands with reconciliation in slice 3c, where `DurableSpool` gains that context. This slice makes provenance mandatory and non-bypassable at the function boundary.

Gate: `make quality`, `test-coverage`, `package` pass — 3261 tests; `reader.py` 98% branch; line 97.89%, branch 96.71%. DCO-clean single commit.

PR #42, opened 2026-07-25
feat(spooling): spool reconciliation scan (W03 slice 3c)

that1guy15:
## W03 slice 3c — spool reconciliation scan

Closes the commit-uncertain gap from slice 3a. Startup and every writer acquisition must reconcile the durable spool with the SQLite ledger (ADR 0004, plan §4.3): a commit publishes the segment file and *then* records its ledger row, so a crash between those steps leaves a durably-published but unrecorded **orphan**.

### What it does
`SpoolReconciler.reconcile()` holds the **exclusive** commit barrier and scans `<spool_root>/pending` against `_segments`:
- **registers each valid orphan** (a trusted, provenance-verified file with no ledger row) — reconstructed **only from the durable header**, never from newer config, with a reconstructed day-start `committed_at` and `origin='reconciled'` so the ledger never records a false precise commit time;
- **reports** every ledger row whose file is missing, every present file that fails trusted validation or disagrees with its row, and every foreign file name;
- returns a `ReconciliationReport(registered, anomalies, healthy, scanned)`. It never deletes, quarantines, or moves a file.

Orphans are registered **only** through `read_trusted_segment` (slice 3b), so a canonical-but-forged or foreign-installation record is never registered as durable state (regression test included).

Migration 3 adds an append-only `origin` column (`DEFAULT 'committed'`); a normal commit omits it, so `commit.py` is unchanged.

### Trust & safety
- All mutation is inside `barrier.exclusive()`; the ledger snapshot and every INSERT share that scope, and committers take the shared side, so no committer can slip a row in between.
- Every failure is a fixed `MH_SPOOL_*` raised outside any handler: registration failure → `MH_SPOOL_RECONCILE`, ledger-read failure → `MH_SPOOL_LEDGER` (both catch `sqlite3.Error` **and** `StateError`).

### Pre-push adversarial review
Confirmed provenance, config-independence, exclusive-barrier mutation, and the fixed-error boundary. It caught a **P1**, now fixed:

> `datetime.strptime('%Y-%m-%d')` accepts a **space-padded** day (`"2026-07- 5"`). Registering it would write a `committed_at` the schema's length-only CHECK admits but the ledger reader's stricter `_STAMP` regex rejects — a **poison row** that fail-closes every later `list_segments`/`read_segment`.

Fix: strict ASCII day regex + a defensive register-time guard; regression proves the ledger stays fully readable. Two P3 notes also addressed (bounded the untrusted day echoed into a report; documented the convergent mid-scan-abort behavior).

### Gate
23 reconciler tests; `reconcile.py` **100% branch**. `make quality`, `test-coverage`, `package` pass — **3284 tests**; line 97.92%, branch 96.75%. DCO-clean single commit.

### Next
Quarantine (move corrupt/foreign files to `quarantine/`), stale-`tmp/` recovery, anomaly persistence, then torn-frame recovery and `spool verify`.

### DIFF
--- a/src/milhouse/spooling/__init__.py
+++ b/src/milhouse/spooling/__init__.py
@@ -2,17 +2,21 @@
 
 from __future__ import annotations
 
-from milhouse.spooling.commit import (
-    DurableSpool,
-    ExporterDelivery,
-    SegmentRecord,
-)
+from milhouse.spooling.commit import DurableSpool
 from milhouse.spooling.errors import SpoolError
+from milhouse.spooling.ledger import ExporterDelivery, SegmentRecord
 from milhouse.spooling.reader import (
     ParsedSegment,
     parse_segment_bytes,
     read_trusted_segment,
     read_untrusted_segment,
+)
+from milhouse.spooling.reconcile import (
+    OrphanRegistration,
+    ReconciliationReport,
+    SegmentAnomaly,
+    SpoolReconciler,
+    run_reconciliation_scan,
 )
 from milhouse.spooling.segment import (
     SegmentHeaderV1,
@@ -30,16 +34,21 @@ from milhouse.spooling.writer import (
 __all__ = [
     "DurableSpool",
     "ExporterDelivery",
+    "OrphanRegistration",
     "ParsedSegment",
+    "ReconciliationReport",
+    "SegmentAnomaly",
     "SegmentHeaderV1",
     "SegmentRecord",
     "SpoolError",
     "SpoolFrameV1",
+    "SpoolReconciler",
     "build_segment_bytes",
     "parse_segment_bytes",
     "publish_segment_bytes",
     "read_trusted_segment",
     "read_untrusted_segment",
+    "run_reconciliation_scan",
     "spool_content_sha256",
     "spool_frame_line",
     "spool_segment_header_line",
--- a/src/milhouse/spooling/commit.py
+++ b/src/milhouse/spooling/commit.py
@@ -1,43 +1,49 @@
-"""Durable spool commit and segment ledger, bound to one control plane (W03 slice 3a).
+"""Durable spool commit, bound to one control plane with mandatory recovery (W03 slices 3a/3c).
 
-A :class:`DurableSpool` binds one control database, one commit barrier, and one spool root that all
-live under the same canonical state root, so a caller cannot publish under a different barrier than
-maintenance holds. Per ADR 0004 and plan section 4.7 a commit is a deliberate, authorized,
-reconciled operation: it authorizes the ``local_spool`` and ``local_sqlite`` egress surfaces
-(restricted input is rejected before any mutation), derives the ``YYYY-MM-DD`` partition from the
-commit instant, then, while holding the shared commit barrier, creates the partition directory,
-atomically publishes the exact segment bytes (write, flush, fsync, no-replace rename, parent fsync),
-and inserts the segment ledger row plus one row per required exporter in a single SQLite
-transaction. A crash or ledger failure after publication leaves a durable orphan and is reported as
+A :class:`DurableSpool` binds one control database, one commit barrier, one spool root, and the
+installation identity, all under the same canonical state root, so a caller cannot publish under a
+different barrier than maintenance holds. Acquiring the store is a writer acquisition: per ADR 0004
+and plan sections 3.4/4.3 it performs mandatory spool/ledger reconciliation under the exclusive
+barrier before any commit is possible, so a durably-published but unrecorded orphan is registered
+before this writer proceeds and recovery is never opt-in.
+
+Per ADR 0004 and plan section 4.7 a commit is a deliberate, authorized, reconciled operation: it
+authorizes the ``local_spool`` and ``local_sqlite`` egress surfaces (restricted input is rejected
+before any mutation), derives the ``YYYY-MM-DD`` partition from the commit instant, then, while
+holding the shared commit barrier, creates the partition directory, atomically publishes the exact
+segment bytes (write, flush, fsync, no-replace rename, parent fsync), and inserts the segment ledger
+row (``origin = 'committed'``) plus one row per required exporter in a single SQLite transaction. A
+crash or ledger failure after publication leaves a durable orphan and is reported as the fixed
 commit-uncertain ``MH_SPOOL_COMMIT``, never as success. The ledger preserves every immutable header
-field so recovery never infers policy from newer configuration, and every read is semantically
-validated so a corrupt or foreign row fails closed with ``MH_SPOOL_LEDGER``. Every failure raises a
-fixed ``MH_SPOOL_*`` error outside the handler.
+field so recovery never infers policy from newer configuration, and every read is validated so a
+corrupt row fails closed with ``MH_SPOOL_LEDGER``. Every failure raises a fixed ``MH_SPOOL_*`` error
+outside the handler.
 """
 
 from __future__ import annotations
 
 import hashlib
 import os
-import re
 import sqlite3
 import stat
-from dataclasses import dataclass
 from datetime import datetime
 from pathlib import Path
-from typing import Any, NoReturn
+from typing import TYPE_CHECKING, NoReturn
 
 from milhouse.config.filesystem import lexical_absolute_path
 from milhouse.core.clock import TimeError, format_timestamp
-from milhouse.privacy import (
-    EgressDisposition,
-    EgressSurface,
-    PrivacyError,
-    require_egress,
-)
 from milhouse.spooling.errors import SpoolError
+from milhouse.spooling.ledger import (
+    ORIGIN_COMMITTED,
+    ExporterDelivery,
+    SegmentRecord,
+    authorize_local_persistence,
+    insert_segment_row,
+    list_segment_records,
+    read_segment_record,
+)
+from milhouse.spooling.reader import INSTALLATION_ID_PATTERN
 from milhouse.spooling.segment import (
-    EXPORTER_ID_PATTERN,
     FRAME_VERSION,
     SCHEMA_VERSION,
     SegmentHeaderV1,
@@ -48,35 +54,15 @@ from milhouse.state.barrier import GlobalCommitBarrier
 from milhouse.state.database import ControlDatabase
 from milhouse.state.errors import StateError
 
+if TYPE_CHECKING:
+    from milhouse.spooling.reconcile import ReconciliationReport
+
 _SPOOL = "spool"
 _BARRIER_NAME = "commit.lock"
 _PENDING = "pending"
 _DIR_MODE = 0o700
 _DELIVERY_PENDING = "pending"
 _DAY_LENGTH = 10
-_SCOPES = ("installation", "target")
-_ALLOWED_PRIVACY = ("public", "internal", "sensitive")
-_DELIVERY_STATES = ("pending", "delivered", "failed")
-_SHA256_HEX = re.compile(r"[0-9a-f]{64}", flags=re.ASCII)
-_STAMP = re.compile(
-    r"[0-9]{4}-[0-9]{2}-[0-9]{2}T[0-9]{2}:[0-9]{2}:[0-9]{2}\.[0-9]{3}Z", flags=re.ASCII
-)
-_SEGMENT_COLUMNS = (
-    "batch_id",
-    "day",
-    "schema_version",
-    "frame_version",
-    "config_generation",
-    "scope",
-    "target_id",
-    "privacy_class",
-    "retention_days",
-    "record_count",
-    "content_sha256",
-    "byte_size",
-    "file_sha256",
-    "committed_at",
-)
 
 
 def _fail(code: str, message: str) -> NoReturn:
@@ -89,51 +75,6 @@ def _current_uid() -> int:
     if geteuid is None:  # pragma: no cover - Milhouse supports only POSIX hosts
         _fail("MH_SPOOL_UNSUPPORTED", "the spool requires a POSIX ownership model")
     return int(geteuid())
-
-
-@dataclass(frozen=True, slots=True)
-class ExporterDelivery:
-    """The delivery state of one required exporter for a committed segment."""
-
-    exporter_id: str
-    delivery_status: str
-
-
-@dataclass(frozen=True, slots=True)
-class SegmentRecord:
-    """A committed segment's immutable ledger row plus its per-exporter delivery rows."""
-
-    batch_id: str
-    day: str
-    schema_version: int
-    frame_version: int
-    config_generation: str
-    scope: str
-    target_id: str | None
-    privacy_class: str
-    retention_days: int
-    record_count: int
-    content_sha256: str
-    byte_size: int
-    file_sha256: str
-    committed_at: str
-    exporters: tuple[ExporterDelivery, ...]
-
-
-def _authorize(privacy_class: str) -> None:
-    denied = False
-    dispositions: tuple[EgressDisposition, ...] = ()
-    try:
-        dispositions = tuple(
-            require_egress(surface=surface, privacy_class=privacy_class)  # type: ignore[arg-type]
-            for surface in (EgressSurface.LOCAL_SPOOL, EgressSurface.LOCAL_SQLITE)
-        )
-    except PrivacyError:
-        denied = True
-    if denied or any(d is not EgressDisposition.REDACTED_RECORD for d in dispositions):
-        _fail(
-            "MH_SPOOL_EGRESS", "the local persistence surfaces do not authorize this privacy class"
-        )
 
 
 def _committed_stamp(committed_at: datetime) -> str:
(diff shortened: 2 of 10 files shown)


that1guy15:
Independent gate review — **fail** at `b4e99c9a809eb871fe72d638fa9f7ea39c609dd9`.

The trusted-reader use, header-only orphan reconstruction, exclusive barrier around explicit reconciliation, strict partition-day validation, transaction rollback, and fixed error normalization are good foundations. The focused spool suites pass locally (`124 passed`), the diff is clean, and all Required CI checks are green. The following recovery-authority defects remain merge-blocking.

[P1] Reconciliation is not mandatory on any writer path  
Action: Change  
Gate/contract: Plan §3.4 and §4.3 require startup **and every writer acquisition** to reconcile the spool and ledger before proceeding; ADR 0004 makes that mandatory recovery step the bridge across file publication and ledger commit.  
Evidence: This PR exports `SpoolReconciler.reconcile()`, but there is no production call site. `DurableSpool.commit_segment` remains unchanged: it authorizes, builds, takes the shared barrier, publishes, and inserts its own row without invoking reconciliation (`commit.py:324-363`). A direct probe published valid orphan `orphan-1`, then committed `later-2`; the ledger afterward was only `[('later-2',)]`. The orphan remained unregistered while a later writer proceeded. This contradicts the module/PR claim that the commit-uncertain gap is closed.  
Impact: Recovery remains opt-in. A process can continue durable state transitions, cursor/derivation work, or a retry while an earlier durable segment is absent from the ledger; retrying the orphan’s batch can also collide with its existing file instead of converging. This is the locked recovery-contract violation the slice is meant to eliminate.  
Exact correction: Put reconciliation behind the single writer-facing acquisition/commit API rather than requiring callers to remember a separate object. Require the installation identity in that bound store/runtime and perform mandatory reconciliation before a writer can publish. Preserve a lock handoff that cannot allow another writer to create a new orphan between reconciliation and the current commit; holding the exclusive barrier through reconcile+commit is a correct initial implementation if a safe atomic downgrade/generation protocol is not yet available. Add the startup hook when the runtime entrypoint exists, and narrow the PR claim if that hook belongs to a named later slice.  
Required regression evidence: Publish a valid orphan, invoke only the normal writer API, and prove the orphan row exists before the later segment is published/recorded. Add same-batch retry, concurrent writer, crash-after-rename, and no-bypass call-graph tests.

[P1] Ledger/file agreement checks only day and whole-file digest  
Action: Change  
Gate/contract: Plan §4.3 says SQLite records matching immutable header metadata plus delivery state and that verification checks ledger agreement; the PR claims every present file that disagrees with its row is reported.  
Evidence: `_ledger_index` selects only `batch_id, day, file_sha256` and the healthy decision compares only name/header batch ID, day, and file SHA (`reconcile.py:248-260, 282-295`). It never reads or compares schema/frame version, config generation, scope/target, privacy class, retention, record count, content digest, byte size, or required exporter identities/status rows. A direct probe changed a committed row to `retention_days=1`, `privacy_class='public'`, `record_count=0`, deleted every exporter row, and then reconciled. The result was `ReconciliationReport(... anomalies=(), healthy=1 ...)`; the corrupted values and empty exporter set remained.  
Impact: Startup can certify corrupted policy and delivery state as healthy. A shortened retention can later cause premature deletion, and missing exporter rows can suppress required delivery—both acknowledged-record loss paths. Classification/scope disagreement also makes the control plane lie about durable content.  
Exact correction: Build the ledger index from the fully semantically validated segment rows plus exporter rows (and origin), then compare every immutable header field, schema/frame versions, byte size/file digest, and the exact required-exporter identity set. Validate delivery statuses even though their mutable values are not header fields. Any malformed row or mismatch must produce a fixed unhealthy anomaly/error, never increment `healthy`.  
Required regression evidence: Independently mutate every ledger/header field and byte size; add/remove/change exporter rows; corrupt origin/status; prove each case is non-healthy and no recovery mutation occurs. Include a complete matching-row positive case.

[P1] Conflicting duplicate orphan batches are registered arbitrarily  
Action: Change  
Gate/contract: Plan §3.4 requires duplicate identity with conflicting content to be quarantined/high-severity, and the spool is the replay authority; recovery must not choose one conflicting durable history by directory sort order.  
Evidence: Registration happens during the first pass before the scan knows whether another candidate has the same batch ID (`reconcile.py:184-220, 262-271`). The test at `test_spool_reconcile.py:391-402` explicitly expects one duplicate to be registered. A stronger probe placed two provenance-valid `batch-1` files on different days with different durable exporter policies. Reconciliation registered the lexicographically earlier day with exporter `alpha` and merely reported the later `beta` file as duplicate.  
Impact: Recovery commits an arbitrary policy/content history to SQLite while a conflicting durable segment exists. That can omit records or required delivery from the other file and turns a detectable conflict into replay corruption.  
Exact correction: Inventory/bound the candidate set before any orphan insert. If a batch ID appears at more than one path, register none—even when bytes happen to match—until an explicit, audited conflict/quarantine rule resolves the ambiguity. Report a fixed conflict anomaly for all candidates; never let sort order select authority.  
Required regression evidence: Same batch in two days with (a) identical bytes and (b) different frames/exporters must produce zero inserted segment/exporter rows and a deterministic conflict result. Re-running must remain mutation-free.

[P2] Filesystem enumeration is unbounded and unsafe names enter the report verbatim  
Action: Change  
Gate/contract: Plan §12 requires every input to be bounded and treated as untrusted; privacy rules forbid raw paths/identifiers from operational surfaces; G03 requires secured `0700` directories. Reconciliation holds the exclusive barrier, so enumeration bounds are availability-critical.  
Evidence: `_list_dir` materializes and sorts all names with unbounded `os.listdir`, and the scan accumulates an unbounded anomaly per entry (`reconcile.py:131-141, 184-230`). It does not securely validate the spool/pending/day directory descriptors or modes before listing. For an invalid day, slicing to 64 characters still returns raw controls/PII (`reconcile.py:198-203`): a directory named `CANARY_SECRET\nforged-field` was returned unchanged in `SegmentAnomaly.day`.  
Impact: A large/corrupt directory can exhaust memory or hold the exclusive barrier indefinitely, blocking every writer. Raw hostile names can inject line breaks or private path text into later health/log/CLI serialization. Loose or replaced directory state is not itself reported reliably.  
Exact correction: Enumerate descriptor-relative with no-follow secured directory handles; validate owned exact `0700` spool/pending/day directories and stable identities. Apply explicit maximum day, entry, anomaly, name-byte, and total-scan bounds before materializing/sorting; fail/report a fixed limit code. Never copy an invalid filesystem name into `day`; use a fixed reason plus an approved keyed fingerprint if correlation is needed.  
Required regression evidence: World-writable/symlinked/replaced directories, over-limit day/entry/anomaly corpora, and control/newline/Unicode/PII canary names must fail safely; canaries must be absent from report repr/serialization, exception, context, and traceback.

[P2] Orphan registration bypasses the common persistence authorization guard  
Action: Change  
Gate/contract: ADR 0007 and plan §12 require redaction/classification and the binding `require_egress` matrix before **every** persistence surface. Normal commit enforces `LOCAL_SPOOL` and `LOCAL_SQLITE` via `_authorize` (`commit.py:123-135`).  
Evidence: `SpoolReconciler._register` inserts the segment and exporter rows directly (`reconcile.py:297-340`) and imports/calls no egress authorization. Trusted structural/provenance validation is not the same control as surface authorization.  
Impact: Recovery becomes a persistence path that bypasses the single fail-closed policy boundary. Today’s matrix admits the canonical non-restricted classes, but the invariant is no longer centrally enforced and a denial/future tightened policy cannot stop SQLite registration.  
Exact correction: Reuse one shared spool/SQLite authorization helper and require the expected `REDACTED_RECORD` disposition before beginning the registration transaction. Convert denial to fixed `MH_SPOOL_EGRESS` and leave both tables unchanged.  
Required regression evidence: Spy on both exact surfaces for a recovered orphan; force denial/unexpected disposition and prove zero segment/exporter rows, fixed safe error, and no rejected value leakage.

[P2] Reconciled timestamp provenance is stored but hidden from the ledger API  
Action: Change  
Gate/contract: The PR adds `origin='reconciled'` specifically so a synthesized day-start timestamp is not presented as a precise original commit time.  
Evidence: Migration 3 stores `origin`, but `commit.py`’s `_SEGMENT_COLUMNS`, `SegmentRecord`, `_validated_segment`, `read_segment`, and `list_segments` omit it (`commit.py:64-120, 226-273, 365-407`). Consumers therefore receive `2026-07-24T00:00:00.000Z` with no way to distinguish it from an exact normal commit.  
Impact: The public control-plane API still asserts false precision and downstream ordering/audit/retention code cannot apply different semantics to reconstructed time. The new column does not achieve its stated truthfulness purpose outside raw SQL.  
Exact correction: Add a validated origin/precision field to `SegmentRecord` and every read/list query, or model recovered time explicitly as unknown-with-day rather than a fabricated exact timestamp. Ensure downstream code cannot treat a reconstructed stamp as exact by accident.  
Required regression evidence: Normal commits surface `committed` with exact time; recovered rows surface `reconciled` (or an explicit day-only value); invalid origin fails closed; read/list round trips preserve the distinction.

Missing or externally pending gate evidence: G03 remains pending by design; quarantine, stale-temp/torn-tail recovery, persisted health, replay/prune, and full crash/concurrency evidence are later work. They are not claimed passed here.

Verdict: **fail**.

that1guy15:
## Gate-review defects addressed — force-pushed at `099067f`

Thanks for the thorough review — all six defects are fixed by a substantial rework, and a second adversarial review confirmed the fixes and surfaced three P3s (also addressed).

**[P1] Reconciliation was not mandatory on any writer path.** Acquiring the writer API is now the mandatory recovery step: `DurableSpool` requires the `installation_id` and runs `run_reconciliation_scan` under the **exclusive** barrier in `__init__`, before any commit is possible. A bad binding still fails before the scan. Regression: publish an orphan, acquire a fresh writer via the normal API, and prove the orphan row exists before the later segment is committed; construction-barrier-failure and no-bypass tests included.

**[P1] Agreement checked only day + whole-file digest.** The scan now loads **fully validated** ledger records and `_agrees` compares every file-attestable immutable field — schema/frame version, config_generation, scope, target_id, privacy_class, retention_days, record_count, content_sha256, byte_size, file_sha256 — and the exact required-exporter identity set. A malformed row is a per-row `corrupt_ledger` anomaly (never healthy, never aborts the scan). Ten parametrized field mutations + exporter add/remove + malformed-row + full-match cases.

**[P1] Conflicting duplicate orphans registered arbitrarily.** Two-pass: the full candidate inventory is counted first, and any batch id at more than one path registers **none** (conflict anomaly for each), even when bytes match — a committed batch that gains a rogue duplicate is also de-certified. Identical-bytes and differing-bytes cases prove zero inserts and a mutation-free re-run.

**[P2] Unbounded/unsafe enumeration; raw names in the report.** `_secure_dir_names` opens each dir `O_DIRECTORY|O_NOFOLLOW`, validates owned exact-0700, and caps entries before materializing; days/entries/total/anomalies are all bounded. Every filesystem-sourced name in an anomaly is a validated identifier or a `sha256:` fingerprint — a `CANARY_SECRET\n…` name is fingerprinted, not echoed. Symlink/world-writable dir and all four bound tests included.

**[P2] Registration bypassed egress authorization.** `_register` is preceded by `authorize_local_persistence(header.privacy_class)` (the shared `local_spool`+`local_sqlite` boundary); a denial records `corrupt_orphan` and leaves both tables unchanged (spied test).

**[P2] Origin stored but hidden.** `SegmentRecord.origin` is validated and surfaced through `read_segment`/`list_segments`; a reconciled row (`origin='reconciled'`, reconstructed day-start `committed_at`) is distinguishable from a committed one, and an invalid origin fails closed.

**Re-review P3s, fixed:** a barrier-acquisition `StateError` now maps to the fixed `MH_SPOOL_BARRIER` on all three writer paths (so only `MH_SPOOL_*` ever escapes); a malformed ledger `batch_id` is fingerprinted, not echoed; and the file-attestable scope of the agreement check is documented.

**Architecture note:** the shared `_segments` primitives moved to a leaf `spooling/ledger.py` that both commit and reconcile import, so neither imports the other (commit imports the reconcile scan lazily in `__init__`).

Gate: `make quality`, `test-coverage`, `package` pass — 3311 tests; reconcile.py 100% branch, ledger.py 99%, commit.py 96%; line 97.95%, branch 96.78%. DCO-clean single commit.

that1guy15:
## Resolution matrix — gate-review findings → commits, regressions, evidence

All six findings were resolved in the rework at branch head `099067f`, squash-merged as **`da1d037`**. A six-auditor evidence audit then verified every *required regression* from the review against merged main; the 14 partial/missing regressions it found are added (test-only, all passing against the merged implementation unchanged) in follow-up **PR #43** (`claude/w03-reconcile-evidence`, commits `a09f9c8`…`6254dbc`), left unmerged for independent re-review.

| # | Finding | Fix (in `da1d037`) | Regression evidence (merged + PR #43) |
|---|---------|--------------------|----------------------------------------|
| 1 | P1 — reconciliation not mandatory on any writer path | `DurableSpool` requires `installation_id`; `__init__` runs `run_reconciliation_scan` under `barrier.exclusive()` before any commit is possible; `SpoolReconciler` remains the explicit startup entry; W05 runtime startup will construct the store (integration point documented) | merged: orphan registered by acquisition alone before a later commit; acquisition barrier-failure → `MH_SPOOL_BARRIER`. PR #43 `a09f9c8`: crashed-commit same-batch retry **converges** (`MH_SPOOL_EXISTS` + ledger shows `origin='reconciled'`, exactly 1 row); spy pins acquisition=**exclusive**/commit=shared; second writer blocks on a held exclusive barrier then reconciles; `__all__` surface pinned — no exported symbol both publishes and registers without acquisition |
| 2 | P1 — agreement checked only day + whole-file digest | `_ledger_index` loads fully validated `SegmentRecord`s (validated_segment + load_exporters + origin); `_agrees` compares every file-attestable immutable field, both digests, byte size, exact exporter identity set; malformed rows → per-row `corrupt_ledger`, never healthy | merged: 10 parametrized field mutations + sole-exporter deletion + malformed-row + full-match positive. PR #43 `ab42a1d`: exporter **add/rename/remove-one-of-several** each de-certify; invalid `delivery_status` rejected by CHECK **and** by the validator without the CHECK |
| 3 | P1 — conflicting duplicates registered by sort order | Two-pass `Counter` inventory before any insert; batch at >1 path registers **none** (identical bytes included); committed+rogue-duplicate de-certified | merged: identical/different-exporter variants, zero rows, mutation-free re-run, committed+duplicate → healthy 0. PR #43 `5d96855`: **different-frames** variant (different source events/record ids) |
| 4 | P2 — unbounded/unsafe enumeration; raw names in reports | `_secure_dir_names`: `O_DIRECTORY\|O_NOFOLLOW` descriptor-relative, owner+exact-0700 fstat, entry cap before materializing; day/entry/total/anomaly bounds; `_fingerprint`/`_safe_id` — validated identifiers or `sha256:` fingerprints only | merged: world-writable pending/day, symlinked day, all four bounds, secret-canary day name, injected ledger batch_id. PR #43 `7d1af2f`: scan-stop pinned as the *only* outcome; ENOTDIR branch; 246-char name + 300-char ledger id fingerprinted; newline/control-byte canary absent from all reprs |
| 5 | P2 — recovery bypassed egress authorization | Shared `ledger.authorize_local_persistence` (`LOCAL_SPOOL`+`LOCAL_SQLITE`, `REDACTED_RECORD` required) called in `_reconcile_orphan` before the registration transaction; denial → `corrupt_orphan`/`MH_SPOOL_EGRESS`, both tables untouched | merged: forced denial writes zero rows; shared-helper admit/deny tests. PR #43 `901542f`: spy proves the **exact surfaces and order**; unexpected non-`REDACTED_RECORD` disposition denies with zero rows; denial error surface carries no rejected value |
| 6 | P2 — origin stored but hidden from the ledger API | `SegmentRecord.origin` validated (`committed`/`reconciled`) and surfaced by `read_segment`/`list_segments`; commit writes `committed`, recovery writes `reconciled` with the reconstructed day-start stamp | merged: list_segments distinguishes both; invalid origin → `MH_SPOOL_LEDGER`. PR #43 `6254dbc`: `read_segment` preserves the distinction independently; origin CHECK rejects unknown values; origin-omitting insert defaults to `committed` through full validation |

**Commands and results** (merged main, and identically at PR #43 head `6254dbc`):
`pytest` focused spool/state/egress suites → 251 passed · full `tests/unit tests/integration` → 3330 passed · `make test-coverage` → line 97.96% / branch 96.81%, 36 critical files (reconcile.py 100% branch, ledger.py 99%, commit.py 96%) · `make quality`, `build`, `package-check` → pass · `check_dco` → 6/6 signed · CI at `099067f`: all 20 checks green.

**Residual risk / deferred (tracked for later slices):**
- Quarantine: reconcile *reports* corrupt/foreign/conflict but does not yet move files (slice 3d), and stale-`tmp/` recovery is 3d scope.
- No extended-ACL check on `spool_root` itself (3d).
- `_MAX_TOTAL` truncation can miss a straddling duplicate (>1M files; self-flags via `scan_limit`; re-detected next scan) — P3.
- `_agrees` covers file-attestable fields only: sub-day `committed_at` and `origin` cannot be attested from the durable file (documented at the comparison site).
- Unsafe-owner fixture not testable without root; replaced-directory race immaterial by construction (single-descriptor fstat+scandir; per-file reads re-validate with post-read snapshot).
- Startup integration point: the W05 runtime entrypoint must construct `DurableSpool` (or run the scan under the exclusive barrier) at startup; the writer path itself is non-bypassable now.

Requesting independent re-review of PR #43; it will not be merged without it.

PR #44, opened 2026-07-26
feat(spooling): quarantine corrupt, conflicted, and stale spool files (W03 slice 3d)

that1guy15:
## W03 slice 3d — quarantine and recovery hardening

This PR resolves spool states that slice 3c only reported (plan sections 3.4 and 4.3). It builds on mandatory exclusive reconciliation at writer acquisition and before every commit, and remains a W03 slice rather than G03 acceptance.

### Quarantine behavior

On a complete, recovery-safe pass under the exclusive barrier, reconciliation handles:

- corrupt committed files while retaining their ledger rows as evidence;
- corrupt or egress-denied orphans;
- exact crashed-writer staging temporaries;
- duplicate batch IDs, preserving a copy that exactly agrees with an existing healthy ledger row and quarantining the remaining duplicate or divergent copies.

Every candidate is classified descriptor-relative with no-follow semantics. Quarantine accepts only an owned, ACL-free, exact-0600, single-link regular file with a bounded captured snapshot and content digest. Symlinks, directories, FIFOs/devices/sockets, foreign owners, loose modes, unexpected links, replacements, and namespace drift fail closed with fixed anomalies.

The move is a digest-verified copy, not a hard link. The copy is staged, fsynced, published without replacement, and re-certified through a freshly validated live `spool_root -> quarantine -> day` chain. Source removal is reversible: the exact source is first renamed to an identity-bound private retirement name and its directory is fsynced; both live pending and quarantine chains plus the target bytes are then re-proven before the retirement name is unlinked and fsynced. Uncertain outcomes retain recoverable evidence and block writer handoff when recovery safety cannot be proven.

### Restrictive-umask and interrupted-stage recovery

Pending/day creation and reconciliation now share one descriptor-relative private-directory primitive. It:

- repairs only an existing owned directory whose mode is a safe subset of 0700;
- validates type, owner, exact mode, ACL envelope, and identity before and after opening;
- fsyncs hardened children and accepted parents, and lets the writer re-prove child durability on retry;
- never creates an absent directory during inventory and refuses symlinks, foreign owners, loose/special modes, ACLs, or identity replacement.

[...]

PR #45, opened 2026-07-26
feat(core): structured-log file surface — persist, rotate, recover, retain (W03 slice 6)

that1guy15:
## What this delivers

The plan §4.15 structured-log filesystem surface over the W02 stored wire (slice 6 of W03, batched in parallel with the spool slices per the agreed strategy): `src/milhouse/core/log_store.py` with `StructuredLogStore` (append/rotate/flush/close), `recover_structured_logs`, and `retention_preview`/`retention_apply`, plus a public `require_no_extended_acl` in `config/filesystem.py` (an identical hunk is on the PR #44 branch; whichever merges second is a no-op).

**Layout and access**: `structured-log.current.jsonl`, twenty-digit rotated names, `.structured-log.next.tmp` staging, `structured-log.lock`; every name opened descriptor-relative, no-follow, close-on-exec, validated owner-only 0700/0600 single-link ACL-free. Every failure is a fixed `MH_LOG_*` `LoggingError` raised outside any handler with no filesystem, JSON, or event value in args/cause/context/traceback.

**Open discipline**: the whole active segment validates at open. Completable crash artifacts (torn tail, staging file, final durable trailer, linked active segment) fail closed `MH_LOG_RECOVERY_REQUIRED`; sequence conflicts (`MH_LOG_SEQUENCE`) and data past a trailer (`MH_LOG_AMBIGUOUS`) are not crash artifacts and stay unrepairable. Surviving rotated names bound the active sequence from below only, so retention pruning never wedges reopen; the live store refuses to create the rotation that would cross `MAX_ROTATIONS`, so maintenance always keeps its API escape.

**Single live appender**: append and rotation revalidate the cached descriptor, link count, and byte size under the log lock and fail stop with `MH_LOG_CONFLICT` before writing when `current` changed beneath the store — cooperating stores can never corrupt a closed segment or silently lose an event. Any failed mutation fails stop (`MH_LOG_CLOSED` afterwards); non-mutating refusals (lock contention, egress refusal) leave the store usable. The docstrings state exactly what `stat` can and cannot distinguish.

**Recovery** (exclusive-hold-gated, lock order barrier → log lock → descriptors): truncates torn incomplete tails, removes crashed staging files, and **completes interrupted rotations** — only when the durable final trailer authenticates against the exact bytes it certifies (canonical form, sequence, count, digest); the sequence-conflict invariant is enforced **before any mutation in both branches**, so a planted trailer can never re-mint a pruned rotation's sequence; squatters, forged-sequence trailers, and foreign extra links are refused in place with retry convergence.

**Retention**: per-segment validation — a corrupt rotation is reported `refused` with its fixed code and left in place, never deleted and never blocking validated expired rotations from pruning (`RetentionReport(segments, refused)`). Effective deadline = captured trailer deadline shortened, never extended, by the current policy.

## Adversarial history (pre-push, all fixed here with reproductions as regressions)

1. **Pre-push review** (P0 + P1 + P2 + P3): the `max+1` sequence-equality check let a routine retention prune permanently brick reopen with no recovery escape; two live stores silently corrupted a closed segment and lost an event; a transient link failure mid-rotation wedged the store unrecoverably; `SecureFileError` escaped uncoded.
2. **Five-agent verification workflow**: recovery bypassed the sequence-conflict check (a planted trailer could re-mint a pruned sequence — now refused `MH_LOG_SEQUENCE` before any mutation, with only the legitimate same-inode twin exempt); a hostile `PathLike.__fspath__` escape; the `MAX_ROTATIONS` self-brick; recovery/open taxonomy splits; the abort-all retention hazard.

## Evidence

- 82 tests; `log_store.py` **100% line and branch**: full line-grammar refusal matrix (incl. non-canonical re-encodings and an ARABIC-INDIC digit name probe), envelope violations, failure injection at every os boundary normalizing to fixed codes, forged/foreign/stale authority refusals, lock timeout on injected monotonic time, captured-deadline semantics, and every reproduction above as a regression.
- `make quality` / `make test-coverage` (98.05% line, 96.78% branch, 36 critical files) / `make build && package-check` — pass. DCO 1/1.
- Deferred: physical crash-persistence evidence on the supported host remains G03 scope (slice 7); W06 wires `local_log` to this store.

https://claude.ai/code/session_01TNmpeujrFdbhBPM4mPxXvN


that1guy15:
Independent Milhouse gate review — **FAIL**

Exact head reviewed: `5359836fc3631f7f814e73e34a1c686f48ae059a`  
Current base reviewed: `97c137fb0fedfa689ee3184e722fbc7516bdb97b`

I reviewed the exact one-commit head as integrated on the current base. The head is a clean descendant of that base, so no synthetic merge was required.

[P1] Retention never closes an expired active segment, so event bytes can outlive the hard privacy deadline indefinitely  
Action: Change  
Gate/contract: Plan §4.7, §4.15, ADR 0007/0016, and W03/G03 require the 14-day log-class privacy bound and say retention rotates an active segment before removing it.  
Evidence: `src/milhouse/core/log_store.py:1282-1310` previews and removes only names returned by `_list_rotated`; the docstring explicitly says the active segment is never deleted and asks an external caller to rotate it. I created an active segment at 2026-07-26 containing one event, closed the store without rotation, and ran `retention_apply` at +30 days under a genuine exclusive hold. It removed zero segments; `structured-log.current.jsonl` remained and still contained the event line. There is no maintenance-owned path here that rotates the due active segment.  
Impact: A quiet installation can retain operational events forever, violating the locked privacy ceiling and the G03 retention contract.  
Exact correction: Under the exclusive barrier and log lock, authenticate the active segment, rotate it when its captured/current effective deadline or maintenance policy is due, publish a fresh successor with the fixed crash order, and then validate/delete the newly closed expired segment.  
Required regression evidence: Leave an event only in `current`, advance beyond captured expiry, and prove apply rotates then removes the old bytes while leaving a valid fresh current header. Cover policy tightening, empty active segments, and crashes at every rotate/apply boundary.

[P1] Trailer validation does not bind `expires_at` to the header, allowing early deletion  
Action: Change  
Gate/contract: Plan amendment A05, §4.15, and ADR 0016 fix `expires_at = opened_at + retention_days`; retention may tighten current policy but may not accept an arbitrary stored deadline.  
Evidence: `src/milhouse/core/log_store.py:1234-1279` validates trailer shape, canonical bytes, event count, and content digest, but never compares `trailer.expires_at` with `header.opened_at + header.retention_days`. Because the trailer is excluded from `content_sha256`, I changed only a valid rotation's canonical trailer deadline from +14 days to +1 day. At +2 days with the unchanged 14-day policy, `retention_apply` accepted the segment and deleted it.  
Impact: A one-field trailer corruption can erase otherwise valid log history before its captured privacy/retention deadline.  
Exact correction: Compute the canonical captured deadline with bounded safe arithmetic during closed-segment authentication and require exact equality before a segment is eligible for preview or deletion. Normalize overflow to a fixed `MH_LOG_*` refusal.  
Required regression evidence: Plant earlier, later, overflow-adjacent, and non-millisecond trailer deadlines while keeping the content digest valid. Each must be refused and left byte-for-byte intact; the normative A05 deadline vector must pass.

[P1] Live rotation unlinks `current` before the rotated name is durably proven  
Action: Change  
Gate/contract: Plan §4.15 and G03 require recovery across every link/unlink/fsync crash boundary and a fixed order that never loses the only name.  
Evidence: `src/milhouse/core/log_store.py:801-809` links `current` to the rotated name, immediately unlinks `current`, and only then fsyncs the directory. My operation trace was `fsync-file, fsync-file, link, unlink-current, fsync-dir`; there is no directory fsync between link and unlink. The recovery helper correctly uses the safer order, but the live rotation path does not.  
Impact: A crash after unlink but before the directory fsync can persist the unlink without proving the new link durable, losing the only filesystem name for a closed segment.  
Exact correction: After the no-replace link, fsync the directory before unlinking `current`; then unlink and fsync the directory again. Recovery must continue to recognize and complete the legitimate two-link crash state.  
Required regression evidence: Pin `link -> fsync(directory) -> unlink(current) -> fsync(directory)` and simulate a crash/failure after every operation, proving at least one authenticated name survives and recovery converges without sequence reuse.

[P1] Maintenance authority is forgeable on this exact head  
Action: Change  
Gate/contract: Plan §4.15, ADR 0004/0016, and the lock-order contract require recovery and retention to run only under a real live exclusive hold of this state root's global commit barrier.  
Evidence: `src/milhouse/core/log_store.py:446-458` trusts only `ExclusiveHold.active` and its path. On this head, `src/milhouse/state/barrier.py:150-167` exposes `_issue()`, which any constructed token can call. I constructed `ExclusiveHold(<state>/control/commit.lock)`, called `_issue()`, held the real barrier shared, and successfully ran `retention_preview`; no exclusive flock existed. The same token authorizes mutating recovery and retention apply.  
Impact: Maintenance can race durable writers despite the declared global barrier, enabling recovery/deletion during an active commit and invalidating the system's crash and acknowledgement proofs.  
Exact correction: Remove activation capability from the token and bind liveness to barrier-owned state that exists only while `GlobalCommitBarrier.exclusive()` genuinely holds the flock; then rebase this head onto that hardened barrier implementation.  
Required regression evidence: Exercise every callable/assignment on a directly constructed token and attempt recovery, preview, and apply while the real barrier is shared. All must fail `MH_LOG_AUTHORITY` with zero filesystem mutation; genuine, foreign, stale, and cross-root holds must be distinguished.

[P1] Replacing the logs directory after open redirects future writes outside `[paths].logs`  
Action: Change  
Gate/contract: Plan §4.15 requires no directory replacement and descriptor-relative access beneath the exact secured logs namespace.  
Evidence: `src/milhouse/core/log_store.py:140-167` opens and validates the directory once, and the store retains that descriptor. `_require_coherent` at lines 693-718 compares only the cached active inode with the `current` name relative to that descriptor; it never proves the descriptor still names the configured logs path. I opened a store, renamed `logs/` to a displaced path, created a replacement owned `0700` `logs/`, then appended and flushed. The append succeeded, the configured logs directory had no current segment, and the displaced directory received the event.  
Impact: A directory swap silently redirects persisted operational metadata outside the configured path while the API reports success.  
Exact correction: Capture the logs-directory identity and re-open/revalidate the configured parent-to-child binding under the log lock before every mutation and maintenance operation; fail stop before writing when the live path no longer names the held descriptor.  
Required regression evidence: Swap the logs directory before append, rotation, recovery, preview, and apply. Every operation must fail with a fixed code and zero writes/deletes outside the exact live namespace.

[P1] Complete noncanonical and privacy-invalid event lines are accepted as valid history  
Action: Change  
Gate/contract: Plan §4.15/A05 and ADR 0016 require exact CanonicalJSONV1 fields/types/order, catalog-owned names and levels, sorted typed metrics, coded nullable errors, validated nullable fingerprints, and no arbitrary text/path/secret content; complete malformed lines must fail closed.  
Evidence: `src/milhouse/core/log_store.py:357-371` checks only the field set plus `line`, `schema`, `privacy`, and a parseable timestamp. It does not validate canonical encoding, level, name, metric shape/order/types, error/fingerprint types or formats, or the exact millisecond timestamp encoding. I appended a complete JSON line with spaces, `name="../../private/path"`, `error="arbitrary-secret-text"`, numeric fingerprint, `level="BOGUS"`, and `metrics="not-an-array"`; reopening succeeded and reported `event_count == 1`. Rotation/retention can subsequently digest and certify those bytes.  
Impact: Corruption or a planted complete line crosses the stored-schema/privacy boundary and becomes accepted durable log history, contradicting the fail-closed wire contract.  
Exact correction: Implement a full v1 event parser/validator that enforces every A05 scalar, nullability, enum, machine-name, metric, coded-error, fingerprint, bound, ordering, and canonical-byte rule, and require exact round-trip canonical bytes before acceptance.  
Required regression evidence: Mutate each event field independently, including types, enum values, metric ordering/duplicates, arbitrary text/path/secret strings, timestamp precision, key order, and whitespace. Reopen, recovery, and retention must all reject the same complete bad line with fixed value-free errors.

[P1] A short initial-header write is published as a successful segment  
Action: Change  
Gate/contract: Plan §4.15 and G03 require complete bounded lines and recovery at every write/fsync/publish boundary; a successful constructor must not publish malformed state.  
Evidence: `src/milhouse/core/log_store.py:623-664` calls `os.write(descriptor, header_line)` but ignores the returned byte count, then fsyncs and renames the staging file to `current` and initializes in-memory size/digest from the full intended line. I injected a one-byte-short header write. Store construction returned successfully with sequence 1 while `current` lacked its terminating LF; the cached state already disagreed with disk.  
Impact: A permitted short write creates an unrepairable/poisoned first or successor segment while the public operation reports success, making the primary log path unusable.  
Exact correction: Use a complete-write loop or require the exact byte count before fsync/publication; on failure, return a fixed error without publishing `current`, leaving only safely classified staging debris for recovery.  
Required regression evidence: Inject zero, partial, interrupted, and error writes for fresh and post-rotation headers. No malformed current may be published and recovery must converge to one valid successor without sequence reuse.

[P1] Retention validates one inode and later unlinks only its name  
Action: Change  
Gate/contract: Plan §4.15 requires a descriptor-bound unlink of only the validated expired closed segment.  
Evidence: `src/milhouse/core/log_store.py:1202-1231` opens, validates, and closes each rotation. `retention_apply` later reconstructs its name and `src/milhouse/core/log_store.py:1319-1327` unlinks that name without carrying or revalidating the certified identity/snapshot. A replacement between preview and unlink is therefore the object deleted, not the object whose expiry was proven. The log lock excludes cooperating stores but does not make this name/identity assertion true.  
Impact: Retention can delete an unvalidated replacement (including unexpired history) while reporting removal of the previously validated segment.  
Exact correction: Carry the opened file identity/snapshot through the retention decision and re-open no-follow plus compare immediately before unlink; revalidate the directory binding and refuse on any drift.  
Required regression evidence: Replace, rewrite, hard-link, chmod, and ACL-modify an expired candidate after validation but before unlink. Apply must leave the replacement intact, report a fixed refusal/uncertain outcome, and never claim the original sequence removed.

[P2] Contended log-lock acquisition busy-spins for the full timeout  
Action: Change  
Gate/contract: W03's bounded concurrency behavior and the five-second injected-monotonic lock wait should not consume a core while waiting.  
Evidence: `src/milhouse/core/log_store.py:244-258` retries nonblocking `flock` in a tight loop; `_LOCK_POLL_SECONDS = 0.05` is defined but never used. The test advances monotonic time on each call, so it does not expose real-time CPU spinning.  
Impact: Every contending process can peg a CPU for up to five seconds, amplifying load during exactly the maintenance/writer overlap the lock is intended to serialize.  
Exact correction: Add an injected bounded wait/sleeper or an equivalent blocking/poll mechanism that honors cancellation/error semantics and the five-second monotonic deadline.  
Required regression evidence: Assert bounded retry count/wait cadence under contention, fixed timeout behavior, and no busy-loop when monotonic time advances normally.

Test/check evidence:

- Exact integrated head: focused structured-log store suite passed locally, `82 passed`.
- Adversarial reproductions on this exact head confirmed the expired-active retention leak, forged early-expiry deletion, invalid event acceptance, successful short-header publication, forged maintenance hold, logs-directory redirection, and missing fsync between rotation link/unlink.
- Local DCO check passed the PR commit; `git diff --check` passed.
- Hosted Required CI run `30222080343` is fully green, including `test`, `quality`, `dco`, workflow `codeql`, separate GitHub CodeQL, package/artifact smoke, and aggregate `required-ci`.

Residual risk: G03 remains pending, and the required supported-host physical crash evidence is still external to unit-level failure injection. The hosted green suite does not cover the contract violations above. Provenance remains a clean-room W03 implementation claim; no donor reuse is asserted by this PR.

Verdict: **FAIL**.

that1guy15:
Independent Milhouse gate review — **FAIL**

Exact head reviewed: `86c72514d809ae7910755af498766ffd7481e512`  
Live protected-`main` base reviewed: `c6edc65992eceb49332f74af7175b61284be2df6`

GitHub's PR object still reported cached `baseRefOid` `549626e7589ab4f66c87ad09acfe6c20305748a3` after PR #52 advanced `main`. I reconciled the exact head against the live branch tip: `git merge-tree --write-tree c6edc65992eceb49332f74af7175b61284be2df6 86c72514d809ae7910755af498766ffd7481e512` produced clean integrated tree `d572f00b55691c81b1dc7efb34d2c1af7748a101`; the intervening base changes are five disjoint community-documentation files.

[P1] Nested maintenance deadlocks by reacquiring the barrier it must receive already held  
Action: Change  
Gate/contract: Plan §4.15 and ADR 0016 fix the lock order and require existing exclusive barrier authority to pass through rather than be reacquired.  
Evidence: `src/milhouse/core/log_store.py:984-985`, `:1416-1417`, and `:1599-1600` validate the barrier object and then always enter a new `barrier.exclusive()`. `GlobalCommitBarrier.exclusive()` at `src/milhouse/state/barrier.py:198-220` is not reentrant and acquires a second descriptor pair. Calling `retention_preview` inside a genuine `with barrier.exclusive():` did not return within two seconds and had to be terminated (`NESTED_MAINTENANCE_TIMEOUT`).  
Impact: The documented maintenance composition path hangs while holding the global barrier and can block every durable writer behind it.  
Exact correction: Pass one non-forgeable lexical exclusive authority through without a second flock, or make exact-instance reentrancy safe while binding its registry to immutable original main/gate paths and rejecting path mutation, foreign instances, stale contexts, and cross-root reuse.  
Required regression evidence: Run recovery, preview, and apply under a genuine outer exclusive hold without another kernel acquisition; prove shared holders block them, foreign/same-path instances are rejected, and path tampering cannot reuse root A authority for root B.

[P1] The rotation bound can retain an expired active segment indefinitely  
Action: Change  
Gate/contract: Plan §4.7/§4.15 and G03 require the log-class privacy deadline and require retention to rotate an expired active segment before removing it. A refused corrupt rotation must not block validated retention around it.  
Evidence: `src/milhouse/core/log_store.py:1606-1612` rotates an expired active only below `MAX_ROTATIONS`; pruning starts later at lines 1625-1642 and never retries the active rotation. With a bounded one-rotation reproduction, one corrupt/refused rotation at the bound, and an expired current containing an event, `retention_apply` returned `active_expired=True`, `active_rotated=False`, refused the rotation, and retained the current event bytes. Repeated calls cannot change that state while the refused name remains.  
Impact: One corrupt rotation at the bound permanently defeats the hard log retention ceiling for later active events.  
Exact correction: Reclaim eligible closed rotations before deciding whether the due active can rotate, and define a fail-closed recovery/quarantine path that preserves evidence without letting refused entries pin expired active bytes forever. Never return a successful apply while a due active remains retained.  
Required regression evidence: Exercise the real bound with expired, unexpired, and corrupt/refused rotations; one apply must retire the active bytes or return an explicit unhealthy/blocked result, with retry convergence and no sequence reuse.

[P1] A symlinked ancestor redirects the log store outside the configured namespace  
Action: Change  
Gate/contract: Amendment A06, §4.15, ADR 0004, and G03 retain traversal/symlink containment and require a securely resolved, symlink-free `[paths].logs` inside the runtime home.  
Evidence: `src/milhouse/core/log_store.py:127-149` performs lexical normalization and applies `O_NOFOLLOW` only to the leaf; `_require_live_directory` at lines 172-198 repeats the same path open. Passing `<root>/alias/logs` where `alias` was a directory symlink succeeded and published `structured-log.current.jsonl` through the symlink (`SYMLINKED_ANCESTOR_ACCEPTED True`).  
Impact: A bypassed or stale path-resolution caller can redirect creation and later descriptor-relative writes outside the approved runtime namespace.  
Exact correction: Require and revalidate the authenticated runtime-path generation, or open the complete state-root-to-logs chain descriptor-relative with no-follow checks for every component and bind it to the expected state root before publication.  
Required regression evidence: Reject symlinks at every ancestor depth for construction, append, rotation, recovery, preview, and apply; retain deterministic directory-displacement and different-user cases.

[P1] Recovery deletes an unauthenticated occupant of the staging name  
Action: Change  
Gate/contract: Plan §4.15 requires no unsafe foreign matching file and says foreign inodes or ambiguous temporary combinations fail closed; recovery may remove only a classified crash artifact.  
Evidence: `src/milhouse/core/log_store.py:1150-1165` stats `.structured-log.next.tmp`, then unlinks it without opening, checking owner/mode/type/link/ACL, validating bytes, or binding identity. A mode-`0600` file containing non-header data at that name was deleted by genuine exclusive recovery, which returned `removed_staging=True`.  
Impact: Recovery destructively removes an object it has not proven belongs to the log protocol.  
Exact correction: Open no-follow, validate the private-file envelope, classify only allowed complete/torn staging states and expected sequence, carry descriptor identity through deletion, and refuse ambiguous/foreign occupants without mutation.  
Required regression evidence: Plant arbitrary regular files, symlinks, hard links, directories/FIFOs, malformed headers, wrong sequences, and swap races; recovery must preserve them while legitimate short-write/fsync/publish crash artifacts still converge.

Required gate evidence is also failing on this exact head:

- Hosted Required CI run `30457431181`: 4,401 tests passed and 2 skipped, but `test` failed because critical `src/milhouse/core/log_wire.py` branch coverage is 91.18%, below 95%.
- `quality` failed because Ruff 0.15.22 would reformat `src/milhouse/core/log_store.py` and `tests/unit/test_log_event_validation.py`.
- Local and hosted DCO checks reject merge commits `4542b3509c0c6757c0b67b1716b7a755a5ca0163` and `86c72514d809ae7910755af498766ffd7481e512` for missing author-matching `Signed-off-by` trailers.
- Aggregate `required-ci` failed. Compatibility on Python 3.11–3.14, identity portability, package/artifact smoke, gitleaks, audit/license, CodeQL, dependency review, and remaining hosted jobs passed.

Additional local evidence:

- The four changed structured-log suites passed: 135 tests.
- A broader source-path run reached 4,398 passes and 1 skip; its four failures were isolated-subprocess imports caused by reviewing an uninstalled temporary worktree. Hosted installed-environment tests above remove that setup limitation.
- Ruff lint passed; `git diff --check` and the clean current-base merge-tree passed.
- The adversarial reproductions above ran against the exact head. No source, PR state, gate ledger, merge, approval, or publication state was changed.

Residual risk: G03 remains pending. Unit failure injection is not the supported-host physical crash/persistence evidence required before this filesystem surface can pass its gate. Provenance remains a clean-room implementation claim; this PR adds no provenance ledger row.

Review verdict: `fail`.

PR #54, opened 2026-07-29
fix(core,state): remediate the five post-merge D04 findings (structured-log surface)

that1guy15:
Remediates the five P1 defects the post-merge exact-head review found on PR #45's merged tree (defect **D04**); G03 is blocked until these are corrected and independently re-reviewed.

**Fixes (each with adversarial regression evidence in `tests/unit/test_log_d04.py`):**
- **D04-1** — `GlobalCommitBarrier.exclusive()` is now exact-instance/exact-thread reentrant via a thread-aware registry: composed maintenance receiving an already-held barrier reuses the flock instead of deadlocking; a different thread/process still contends on the flock; a same-thread foreign instance and a path mutated mid-hold fail closed.
- **D04-2** — maintenance derives the state root from the barrier's own lock path (not `logs.parent`), so a plan-valid nested `[paths].logs` is authenticated, not rejected.
- **D04-3** — retention reclaims expired closed rotations before rotating a due active segment, then prunes it; if the bound stays pinned it returns an explicit `MH_LOG_RETENTION_BOUND` blocked result — never a silent apply that keeps expired bytes past the ceiling.
- **D04-4** — the state-root→logs chain is opened descriptor-relative with `O_NOFOLLOW` per component, rejecting a symlinked ancestor at any depth for construction, append, rotation, recovery, preview, and apply.
- **D04-5** — recovery opens the staging occupant no-follow/non-blocking, requires the private-file envelope, classifies only a genuine crash artifact (empty / canonical-header-preamble prefix / complete successor header at an unused sequence) before an identity-bound unlink, and refuses foreign or ambiguous occupants without mutation.

An independent exact-head review returned **PASS** (no P0/P1); its one P2 (an over-permissive torn-header prefix) is fixed in the second commit. Local gates green: `test-coverage` (branch 96.81%, all 49 critical files ≥95%), `quality`, `mypy`, `docs-check`, `skill-check`, `secret-scan` (tree + history). Commits are DCO signed-off.

### DIFF
--- a/src/milhouse/core/log_store.py
+++ b/src/milhouse/core/log_store.py
@@ -140,13 +140,18 @@ def _resolve_logs_dir(logs_dir: str | Path) -> Path:
     return resolved
 
 
-def _open_logs_dir(logs_dir: Path) -> int:
-    """Open the logs directory as an owned, exact-0700, ACL-free no-follow descriptor."""
+def _close_fd(descriptor: int) -> None:
+    try:
+        os.close(descriptor)
+    except OSError:  # pragma: no cover - defensive close failure
+        pass
+
+
+def _require_logs_envelope(descriptor: int) -> None:
+    """Require an opened logs directory to be an owned, exact-0700, ACL-free directory."""
 
     unsafe = False
-    descriptor: int | None = None
     try:
-        descriptor = os.open(logs_dir, os.O_RDONLY | _NOFOLLOW | _DIRECTORY | _CLOEXEC)
         info = os.fstat(descriptor)
         if (
             not stat.S_ISDIR(info.st_mode)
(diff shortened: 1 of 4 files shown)


that1guy15:
Independent Milhouse gate review (report-only, exact head) — **PASS**

Exact head reviewed: `51b2b2fc` on base `main` `5624d1c`. Read-only adversarial review of the barrier and log-store diff, verifying each D04 fix closes its finding and hunting for new defects introduced by the remediation.

- **D04-1 (barrier reentrancy)** — CORRECT. Exact-instance/exact-thread reentrant `exclusive()`: nested same-instance reuses the flock; a different thread genuinely contends (blocks, not reentrant); same-thread foreign instance and mid-hold path mutation fail closed. Registry lifecycle leak-free across normal, outer-exception, and nested-exception exits; `exclusive_then_shared`/`shared` unaffected.
- **D04-2 (nested logs)** — CORRECT. State root derived from `barrier.path`; `lexical_absolute_path` normalizes so `is_relative_to` is safe; no cross-root authority grant.
- **D04-3 (rotation bound)** — CORRECT. Prune-then-rotate, then prune the just-closed rotation; explicit `MH_LOG_RETENTION_BOUND` blocked result; accounting sound, no double-prune.
- **D04-4 (symlinked ancestor)** — CORRECT. Per-component `O_NOFOLLOW` walk; `_require_live_directory` re-walks (catches `LoggingError` + `OSError`) for append/rotation/recovery/preview/apply; no fd leak.
- **D04-5 (staging occupant)** — SUBSTANTIALLY FIXED; one **P2** (torn-header prefix accepted arbitrary trailing bytes) — now tightened to require a genuine canonical-header-preamble prefix, with a regression test.

No `__context__`/`__cause__` leaks introduced; no regression in spool/barrier consumers. Overall verdict: **PASS** (no P0/P1). The P2 is fixed in this branch.

PR #67, opened 2026-07-30
fix(w03): re-remediate the D05 re-review findings (4 P1 + 1 P2)

that1guy15:
The independent gate review failed the first D05 remediation (PR #65). This re-remediates every reproduced finding, each with adversarial regression evidence. The ledger was already corrected on `main` by #66.

**This PR is not to be self-merged** — a fresh independent full-SHA review is the acceptance gate (that self-certification was D05 finding 5). Landed only after an independent review passes.

## The four P1s
1. **Frame-snapshot TOCTOU** — `deliver_segment` freezes the caller's frames to one immutable tuple up front and both validates and forwards *that exact tuple*; a `Sequence` yielding different contents on a later read can no longer forward one batch while the digest certified another. Malformed frame objects are rejected. *(test: a mutating sequence is materialized exactly once and the validated snapshot is what's forwarded.)*
2. **Delivery races privacy-expiry retention** — the whole reload → forward → checkpoint runs under a single `barrier.shared()` hold (the CAS no longer re-acquires the barrier, which could deadlock at the writer-preference gate), so `retention_apply` (exclusive) cannot prune mid-delivery. Delivery is also **expiry-aware**: a segment with any expired record is withheld (`expired`), never forwarded — enforcing hard privacy-expiry even before retention runs. `deliver_segment`/`replay_segments` now take an explicit `now`. *(tests: exclusive maintenance is fenced out during an in-flight delivery; an expired segment is withheld.)*
3. **Missing row conflated with delivered** — the delivery-status CAS distinguishes a vanished row (`segment_gone`) from an already-terminal one (`already_delivered`), so a segment pruned out from under a delivery is never certified delivered. *(test: a self-pruning exporter → `segment_gone`.)*
4. **Audit charset ≠ secret rejection** — `record_retention_prune` stores the resource **only as a SHA-256 fingerprint**, so `_audit` can never hold a raw caller string even when a legal opaque id is secret-shaped (an access key is a valid batch-id shape). *(tests: all-alphanumeric credential canaries incl. `AKIAIOSFODNN7EXAMPLE` are fingerprinted, never stored raw.)*

## The P2
- **Post-unlink orphan misclassification** — a descriptor-close failure after a successful unlink now maps to `COMMIT_UNCERTAIN` (the file is gone), not `UNREADABLE`, so retention reports it uncertain rather than re-adopting a non-existent orphan.

## Evidence
Full local gates green: `test` (4602 passed / 1 skipped), `test-coverage` (line 97.95% / branch 96.92%, every touched critical file ≥95%), `quality` (ruff, mypy, docs 74/76, validators). DCO signed.

### DIFF
--- a/src/milhouse/config/filesystem.py
+++ b/src/milhouse/config/filesystem.py
@@ -884,7 +884,12 @@ def remove_regular_file_no_follow(
                     os.close(descriptor)
                 except OSError:
                     if failure is None:
-                        failure = SecureFileError(SecureFileErrorKind.UNREADABLE)
+                        # Reaching the close with no prior failure means the whole body — including
+                        # the unlink — succeeded, so the entry is already gone. A descriptor-close
+                        # failure then leaves durability unconfirmed, never a lingering orphan:
+                        # report commit-uncertain so retention re-inspects rather than asserting the
+                        # file is still present.
+                        failure = SecureFileError(SecureFileErrorKind.COMMIT_UNCERTAIN)
 
     if failure is not None:
         raise failure
--- a/src/milhouse/spooling/exporter.py
+++ b/src/milhouse/spooling/exporter.py
@@ -6,22 +6,34 @@ concrete ClickHouse exporter is W04. The :class:`Exporter` contract is deliberat
 and a ``deliver`` that returns on destination confirmation and raises on any failure.
 
 Delivery is integrity-bound so a terminal ``delivered`` can never be recorded for the wrong data
-(D05): :func:`deliver_segment` binds the barrier to the live control database, reloads the segment's
-authoritative ledger row and rejects a supplied record that does not match it, and requires the
-supplied frames to hash to that row's ``content_sha256`` (with matching batch id and count) before a
-single byte is forwarded — so a caller cannot forward one segment's frames while certifying another.
-It also requires each exporter object to self-identify as the row it delivers, and treats
-the final compare-and-set as authoritative: only a CAS that actually advances the expected row to
-``delivered`` is reported as a new delivery.
+(D05): :func:`deliver_segment` freezes the caller's frames into one immutable tuple, binds the
+barrier to the live control database, reloads the segment's authoritative ledger row and rejects a
+supplied record that does not match it, and requires that exact frozen tuple to hash to the row's
+``content_sha256`` (with matching batch id and count) before a single byte is forwarded — so a
+caller cannot validate one frame set and forward another, nor forward one segment's frames while
+certifying another. It also requires each exporter object to self-identify as the row it delivers.
+
+Delivery is *fenced* and *expiry-aware* (D05 re-review):
+
+* **Fenced.** The whole reload → forward → checkpoint runs under a single ``barrier.shared()`` hold,
+  so privacy-expiry ``retention_apply`` (which prunes under the *exclusive* side) cannot delete a
+  segment mid-delivery. The shared side is held across the destination round-trip; maintenance waits
+  for in-flight deliveries to drain, exactly as a readers-writer lock requires.
+* **Expiry-aware.** A segment with any expired record is withheld — never forwarded — because
+  record-class privacy expiry is a hard upper bound (plan §§4.8-4.9): expired data must never
+  egress, even if retention has not pruned it yet. A mixed-expiry segment waits for audited
+  compaction (slice 5c) rather than leaking its expired records.
 
 The delivery ledger (``_segment_exporters.delivery_status``) is a per-(segment, exporter) state
 machine: ``pending``/``failed`` are retryable, ``delivered`` is terminal. Rule 12 requires the
-exporter checkpoint to advance only after the destination confirms, so delivery is *outside*
-the barrier and, only once ``deliver`` returns, recorded in one shared-barrier compare-and-set that
-never overwrites a ``delivered`` row. A crash between confirmation and the write leaves the row
-retryable, so delivery is at-least-once and exporters MUST be idempotent. Every SQLite or barrier
-failure normalizes to a fixed ``MH_SPOOL_EXPORT`` error, and a misbehaving exporter that raises is
-contained as a failed delivery.
+exporter checkpoint to advance only after the destination confirms, so delivery is recorded, once
+``deliver`` returns, in one compare-and-set that never overwrites a ``delivered`` row. That CAS is
+authoritative: only a CAS that actually advances the expected row is reported as a new delivery, a
+row that is already ``delivered`` is reported ``already_delivered``, and a row that has *vanished*
+(pruned out from under the delivery) is reported ``segment_gone`` — never conflated with a delivery.
+A crash between confirmation and the write leaves the row retryable, so delivery is at-least-once
+and exporters MUST be idempotent. Every SQLite or barrier failure normalizes to a fixed
+``MH_SPOOL_EXPORT`` error, and a misbehaving exporter that raises is contained as a failed delivery.
 """
 
 from __future__ import annotations
@@ -29,8 +41,10 @@ from __future__ import annotations
 import sqlite3
 from collections.abc import Mapping, Sequence
 from dataclasses import dataclass
+from datetime import datetime
 from typing import NoReturn, Protocol, runtime_checkable
 
+from milhouse.core.clock import TimeError, format_timestamp
 from milhouse.spooling.errors import SpoolError
 from milhouse.spooling.ledger import SegmentRecord, read_segment_record
 from milhouse.spooling.segment import (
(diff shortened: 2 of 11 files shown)


that1guy15:
Independent Milhouse gate review — **FAIL**

Exact head reviewed: `4bbad6ff06caaf8b78fc5209faa4c76a57f7ac23`  
Current base reviewed: `bf4f8d1afdae147ac6080458a5d68c1378686dda`  
Integrated merge ref reviewed: `3e763546dc49dde0ecee2c1470bc55353655da78` (parents verified as that base and exact head).

I reviewed all 11 changed files on the exact integrated tree against plan 1.0, W03/G03, ADRs 0004/0007/A06, the current status ledger, architecture, threat model, security, provenance, and the D05 claims. The frame sequence is now frozen once; the bound shared barrier covers reload → forwarding → checkpoint; a vanished delivery row is distinguished; and post-unlink close failure is no longer reported as a lingering orphan. Three P1 contracts remain.

### [P1] Replay forwards after immutable header/ledger policy disagreement

**Action:** Change  
**Gate/contract:** Plan §§4.3 and 3.4 require replay to validate header/frame/ledger agreement before forwarding; ADR 0004 makes the self-describing spool the replay authority; G03 may not certify replay corruption or privacy-policy laundering.  
**Evidence:** `src/milhouse/spooling/replay.py:59-70` compares only file digest, byte size, content digest, and record count. It omits day, schema/frame version, configuration generation, scope/target, privacy/retention class, and required exporters. `src/milhouse/spooling/exporter.py:123-136` repeats the same partial comparison before forwarding. On this exact integrated tree I committed an `internal` segment whose immutable header required `clickhouse`, changed only the live ledger to `public` plus exporter `hosted`, and replay reported `delivered` after calling the `hosted` exporter. The canonical verifier correctly reported `intact=False`, `MH_SPOOL_VERIFY` for the same state.  
**Impact:** A valid file plus semantically drifted/corrupt control rows can redirect its records to a destination the durable header never authorized, launder its privacy classification, and then record a terminal delivery. That breaks replay authority, privacy, and recovery correctness.  
**Exact correction:** Before any exporter call, compare the parsed trusted segment to the freshly reloaded ledger with the canonical full agreement predicate (or an equivalent single source of truth) under the same shared barrier used through checkpointing. Include every file-attestable immutable field and the exact required-exporter set; fail closed before forwarding on any mismatch.  
**Required regression evidence:** Independently perturb each immutable ledger/header field, especially `privacy_class` and `required_exporters`; `verify_spool`, replay, and direct delivery must all refuse, no exporter may be called, and no delivery row may advance.

### [P1] A later exporter can begin after hard privacy expiry

**Action:** Change  
**Gate/contract:** Plan §§4.3 and 4.7 plus ADRs 0004/0007 make record-class expiry a hard upper bound: pending records retry only until expiry and expired data must not egress.  
**Evidence:** `src/milhouse/spooling/exporter.py:283-296` checks the caller-supplied `now` only once before iterating all exporters; `src/milhouse/spooling/exporter.py:296-322` never refreshes the authorization. On this exact tree I used a record expiring two seconds after delivery began, a first exporter that occupied the call for three seconds, and a second exporter. The result was `call_before_expiry=True`, `late_started_after_expiry=True`, outcomes `delivered,delivered`.  
**Impact:** One slow destination can carry the operation across the privacy deadline and a later destination then receives and certifies already-expired records, contradicting this PR's “never forwarded” remediation claim.  
**Exact correction:** Make expiry authorization current per destination attempt rather than a one-time operation snapshot. Inject/read a trusted wall clock immediately before each exporter call and define/enforce a delivery deadline or remaining-time bound so the protocol cannot start or certify egress after the earliest record expiry.  
**Required regression evidence:** Use an injected advancing clock (not wall-clock sleeps) with at least two exporters; after the first advances beyond the earliest expiry, the second must be withheld as `expired`, never called, and remain non-delivered. Cover the exact-boundary case and a mixed-expiry segment.

### [P1] Audit “fingerprints” are public unsalted hashes of secret-shaped input

**Action:** Change  
**Gate/contract:** Plan §4.7 requires secret values to be removed or policy-pseudonymized and explicitly says pseudonyms use installation-keyed HMAC, never a public unsalted hash; ADR 0007 requires privacy-safe SQLite audit metadata.  
**Evidence:** `src/milhouse/state/audit.py:71-81` computes plain `sha256(value)`, and line 167 persists it. The new tests deliberately accept access-key/PAT-shaped values and assert that public digest. A public candidate guess reproduces the stored value exactly. The repository already documents this precise weakness at `src/milhouse/spooling/reconcile.py:248-258`: a bare SHA-256 is dictionary-recoverable and must not represent invalid/untrusted names without keyed pseudonymization.  
**Impact:** The `_audit` table retains stable, cross-installation, offline-guessable derivatives of credential-shaped caller input. Removing the raw spelling is not sufficient under the locked pseudonym contract, so the D05 audit-privacy finding remains unresolved.  
**Exact correction:** Either bind this constructor to a proven canonical internal resource and reject secret-shaped/non-authoritative caller values before persistence, or inject the installation pseudonymizer and store a domain-separated keyed fingerprint. Do not persist public SHA-256 derivatives of potentially secret input.  
**Required regression evidence:** The access-key/PAT canaries must be rejected before insert or stored only as installation-keyed, domain-separated values that differ across keys and cannot be matched by public SHA-256 guessing; raw values, public digests, errors, causes, contexts, and tracebacks must remain absent.

### Test and check evidence

- Exact integrated checkout: merge parents are the reviewed base and head; `git diff --check` passed.
- Targeted remediation suite: **101 passed**.
- Full local test gate: **4,602 passed, 1 skipped** on macOS / Python 3.14.6; 13 pytest cleanup warnings were emitted, with no test failure.
- Coverage gate: **97.95% line / 96.92% branch; 56 critical files passed the ≥95% branch rule**.
- Local quality, docs/link, workflow, skill, private-identifier, tree/history secret scans, and DCO all passed; DCO covered the sole commit.
- Exact-head GitHub checks: **20 completed, 20 successful**, including protected `required-ci`, DCO, both CodeQL checks, gitleaks, dependency review, audit, package/artifact smoke, compatibility, identity portability, quality, and test. No legacy commit statuses exist.
- Green gates do not exercise the three adversarial paths above.

### Residual / pending gate risk

- This is a W03 remediation slice, not complete G03 evidence. The status ledger correctly keeps W03/G03 in progress; full crash/concurrency, slice-5c compaction, supported-host, and remaining gate evidence stay pending.
- No live provider was called; none is required for this W03 protocol review.

**Verdict: FAIL.** Do not merge this exact head. Correct all three P1s, add the regression evidence above, and request a fresh independent review of the new full SHA.

that1guy15:

- Exact head SHA: `4bbad6ff06caaf8b78fc5209faa4c76a57f7ac23`
- Reviewed base SHA: `bf4f8d1afdae147ac6080458a5d68c1378686dda`
- Integrated candidate: the reviewed base is an ancestor of this one-commit head; reviewed tree `b624d5aee708dc645ed3f87cf4959ca650095825`
- Scope: all 11 changed source/test files, the W00-W18 plan, current W03/D05 status, ADRs 0003/0004/0007, architecture, provenance, security, hosted checks, DCO, and adversarial delivery/retention behavior

### Findings

[P1] Delivery can cross the privacy-expiry deadline while preventing retention from acting

Action: Change

Gate/contract: `docs/implementation-plan.md` §§4.3 and 4.7 require record-class privacy expiry to remain a hard upper bound and pending records to remain retryable only until delivery or expiry, whichever comes first; ADR 0007 repeats that privacy bound. This PR specifically claims to remediate the D05 delivery/retention expiry race.

Evidence: `src/milhouse/spooling/exporter.py:280-318` acquires `barrier.shared()`, evaluates expiry once using the caller-supplied call-start `now` at line 283, then invokes the synchronous destination at line 313 and records `delivered` at lines 315-318 without any deadline-aware exporter contract or fresh-clock enforcement. Because the shared hold also excludes `retention_apply`, time can advance past `expires_at` while the destination call is in flight and retention cannot intervene. On this exact head, an adversarial exporter started with `now` 300 ms before the record deadline, waited 450 ms before its simulated send, verified the exclusive barrier was still busy after expiry, and then returned normally. Observed deadline: `2026-07-30T20:42:30.327000+00:00`; send: `2026-07-30T20:42:30.507094+00:00`; result: `delivered`; ledger: `delivered`.

Impact: a slow, rate-limited, or stalled exporter can egress pending record data after its immutable privacy deadline. A long replay can do the same for later segments because one stale `now` is reused across the pass. The new barrier converts the earlier retention race into a time-of-egress expiry violation, so the claimed D05 P1 remediation is incomplete.

Exact correction: make the exporter protocol deadline-aware and enforce the earliest frame `expires_at` across the actual destination operation using a fresh trusted clock and a bounded/cancellable destination timeout. Expiry must be able to prevent the send itself; a check only after `deliver()` returns is too late because the bytes may already have left. Reconcile barrier ownership so maintenance is not indefinitely prevented from enforcing expiry, and do not mark a deadline-crossing attempt delivered.

Required regression evidence: add a deterministic controlled-clock exporter test that begins before expiry and reaches its send point after expiry, proving that zero frames are forwarded, no delivered checkpoint is recorded, and expiry/retention can take authority. Cover the same case during a multi-segment replay whose initial `now` becomes stale.

### Test and check evidence

- Exact toolchain: `uv 0.11.29`, CPython 3.14.6.
- Targeted changed-surface suite: 101 passed.
- `python3 scripts/run_make.py test`: 4,602 passed, 1 skipped.
- `docs-check`: 69 files, 238 links, 74 external probes passed.
- `skill-check`: five canonical skills and discovery aliases valid.
- `secret-scan`: private-identifier, tree, and full-history scans passed after refetching the unfiltered public history.
- `git diff --check`: passed; isolated checkout clean apart from ignored tool environment.
- Local DCO: 1 commit passed.
- Hosted checks on the exact head are green, including `required-ci`, DCO, CodeQL, test, quality, audit, gitleaks, dependency review, package/artifact smoke, compatibility, and identity portability.
- Adversarial expiry-crossing reproduction: **failed the contract** as described above despite the green suite.

### Missing / externally pending gate evidence

- The P1 must be corrected with the deadline-crossing regression evidence above and receive a fresh independent review on the new exact head. No W03/D05 completion or gate advancement is justified by this head.

### Residual risk

The added tests do cover the prior mutable-sequence TOCTOU, missing-row CAS distinction, secret-shaped audit resource, maintenance exclusion, and post-unlink close classification. They do not cover time advancing across a real destination operation, which is the blocking residual privacy risk.

**Verdict: FAIL**

PR #69, opened 2026-07-31
feat(w03): audited restartable compaction of mixed-expiry segments (slice 5c)

that1guy15:
W03 slice 5c. A **mixed-expiry** segment (some records expired, some live) can neither be pruned by retention (it would drop live records) nor delivered (its expired records must not egress). `compact_apply` rewrites it into a NEW segment holding only the unexpired frames — removing the expired ones while **never losing a live record** (ADR 0004 / plan §§4.8-4.9).

## Protocol
Under one `barrier.exclusive()` hold (confirm-token gated, spool-root bound to `<state_root>/spool`), for each mixed segment:
1. Re-read + re-verify the old file against its ledger row.
2. Build the new segment from the surviving frames, with a `batch_id` derived **deterministically** from the survivors' record ids.
3. Publish its fsynced file; **VERIFY** it reads back and agrees with its intended row.
4. In ONE transaction: insert the new row, re-point any source cursor to the new segment, delete the old row, record the old→new lineage audit.
5. Only after that commit, unlink the old file.

## Restartable / convergent
The deterministic new `batch_id` makes a re-run idempotent. A crash before the swap leaves the new file as a re-registerable orphan (reconciliation adopts it, a later pass finishes the retire); a crash before the old-file unlink leaves the old file as an orphan of a still-mixed segment (a later pass finds the new segment already committed and just retires the old). The D05 hard-expiry gate withholds the still-present old mixed segment throughout, so **no expired frame egresses mid-compaction**.

## Surface
- `state/audit.py`: `record_compaction` — writes old→new lineage as two linked rows (`compacted_from`/`compacted_into`), each id stored only as a SHA-256 fingerprint, on the caller's transaction (the `_audit` table needs no migration).
- `spooling/compaction.py`: `compact_apply`, `CompactionResult`, `CompactedSegment`, `SkippedSegment`. The new segment inherits the old policy fields and per-exporter delivery status; old-file removal is the same tri-state (removed/orphaned/uncertain) as retention.
- `Makefile` + `test_makefile_safety`: `compaction.py` added to the 95%-branch critical-coverage floor.

## Evidence
25 adversarial tests: happy path, every skip reason (live / fully-expired / unreadable / disagreeing), all three crash boundaries with reconvergence, verify-before-unlink, publish/build/commit failures leaving the old intact, cursor re-pointing, delivered-status inheritance, audit lineage. Full local gates green — `test` (4628 passed / 1 skipped), `test-coverage` (line 97.97% / branch 96.94%, `compaction.py` 100% branch, 57 critical files), `quality` (ruff, mypy, docs). DCO signed.

Not self-merged — acceptance is the owner's (independent review or direct approval). The ledger will be updated to record slice 5c only on acceptance.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,6 +103,7 @@ test-coverage:
 		--critical 'src/milhouse/state/migrations.py' \
 		--critical 'src/milhouse/state/schema.py' \
 		--critical 'src/milhouse/spooling/commit.py' \
+		--critical 'src/milhouse/spooling/compaction.py' \
 		--critical 'src/milhouse/spooling/errors.py' \
 		--critical 'src/milhouse/spooling/exporter.py' \
 		--critical 'src/milhouse/spooling/ledger.py' \
--- a/src/milhouse/spooling/__init__.py
+++ b/src/milhouse/spooling/__init__.py
@@ -3,6 +3,12 @@
 from __future__ import annotations
 
 from milhouse.spooling.commit import DurableSpool
+from milhouse.spooling.compaction import (
+    CompactedSegment,
+    CompactionResult,
+    SkippedSegment,
+    compact_apply,
+)
 from milhouse.spooling.errors import SpoolError
 from milhouse.spooling.exporter import Exporter, ExporterAttempt, deliver_segment
 from milhouse.spooling.ledger import ExporterDelivery, SegmentRecord
@@ -43,6 +49,8 @@ from milhouse.spooling.writer import (
 )
 
 __all__ = [
+    "CompactedSegment",
+    "CompactionResult",
     "DurableSpool",
     "Exporter",
     "ExporterAttempt",
@@ -60,11 +68,13 @@ __all__ = [
     "SegmentRecord",
     "SegmentRetention",
     "SegmentVerification",
+    "SkippedSegment",
     "SpoolError",
     "SpoolFrameV1",
     "SpoolReconciler",
     "VerifyReport",
     "build_segment_bytes",
+    "compact_apply",
     "deliver_segment",
     "parse_segment_bytes",
     "publish_segment_bytes",
--- a/src/milhouse/spooling/compaction.py
+++ b/src/milhouse/spooling/compaction.py
@@ -71,6 +71,7 @@ _PENDING = "pending"
 _SPOOL = "spool"
 _BARRIER_NAME = "commit.lock"
 _EXISTS_CODE = "MH_SPOOL_EXISTS"
+_DELIVERED = "delivered"
 
 # The compacted segment's batch id is this prefix plus the SHA-256 hex of the surviving records'
 # ordered ids, so it is deterministic (re-runs are idempotent) and cannot collide with a batch id.
@@ -261,6 +262,37 @@ def _remove_old_file(path: Path) -> str:
     return FILE_REMOVED
 
 
+def _is_intended_successor(existing: SegmentRecord, intended: SegmentRecord) -> bool:
+    """Whether an existing row at the derived id is EXACTLY the successor this compaction intends.
+
+    A committed segment's ``batch_id`` is caller-chosen, so the deterministic derived id could be
+    occupied by an unrelated segment (no cryptographic collision needed). Reusing such a row would
+    retire the mixed segment against foreign data and lose the live records, so an existing row is
+    reusable only if every file-attestable identity field — including both digests — and the
+    exporter id set equal the intended successor's. ``origin`` (a crash-re-registered orphan is
+    ``reconciled``) and the per-exporter delivery statuses are deliberately excluded: the former is
+    not identity, and the latter are restored to the intended states during the swap.
+    """
+
+    return (
+        existing.batch_id == intended.batch_id
+        and existing.day == intended.day
+        and existing.schema_version == intended.schema_version
+        and existing.frame_version == intended.frame_version
+        and existing.config_generation == intended.config_generation
+        and existing.scope == intended.scope
+        and existing.target_id == intended.target_id
+        and existing.privacy_class == intended.privacy_class
+        and existing.retention_days == intended.retention_days
+        and existing.record_count == intended.record_count
+        and existing.content_sha256 == intended.content_sha256
+        and existing.byte_size == intended.byte_size
+        and existing.file_sha256 == intended.file_sha256
+        and {e.exporter_id for e in existing.exporters}
+        == {e.exporter_id for e in intended.exporters}
+    )
+
+
 def _swap_ledger(
     database: ControlDatabase,
     old_record: SegmentRecord,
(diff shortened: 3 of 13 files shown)


that1guy15:
## Independent Milhouse gate review — FAIL

- Reviewed head SHA: `6ff9e6f5d58dce2d4efd4073e17b9a3395947fa6`
- Reviewed base SHA: `ba1903a8fd7b3c39ea40aa325b92d966e0c8d064`
- Integrated review commit: `1f307b892bcb8bd2d03b133cf9ff8055b134a0c8` (parents base then head)
- Scope: all 8 changed files plus the authoritative plan/status, ADR 0004, ADR 0007, architecture, security, provenance, spool ledger/reconciliation, retention, exporter, cursor, audit, packaging, and test contracts.

### Findings

[P1] An existing derived batch ID can make compaction delete the only live copy

Action: Change

Gate/contract: Plan §§3.2, 4.3, and 4.7; ADR 0004; G03. Compaction must remove only expired frames, never prune the last recoverable unexpired copy, and verify the intended new segment before retiring the old one.

Evidence: `src/milhouse/spooling/compaction.py:191-196` derives the new ID only from survivor record IDs. At `src/milhouse/spooling/compaction.py:347-360`, any ledger row already using that predictable ID becomes `reference`; `src/milhouse/spooling/compaction.py:362-371` verifies that existing file against its own row, never against the newly built `new_record`, and then deletes the old row/file. On this exact integrated tree I committed a mixed `batch-a`, precommitted an unrelated valid segment at the predictable derived ID, and ran `compact_apply`. It reported `batch-a` compacted, unlinked its file, retained only the unrelated record/exporter, and the intended live record ID disappeared from the spool. This requires no SHA-256 collision.

Impact: acknowledged unexpired data can be lost, a source cursor can be repointed to unrelated bytes, and required exporter policy can be replaced by the unrelated row. This is the G03/SECURITY P1 acknowledged-data-loss class.

Exact correction: make compacted identity bind the old batch plus the exact intended compacted bytes and immutable policy, and never treat “row exists” as convergence by itself. Before retirement, require the existing file and row to equal the intended `new_record` semantics (including content/file digests, policy, exporter set/status authority, day, and origin/recovery state); otherwise fail closed. Make the ledger swap conditional on the exact expected old and new rows.

Required regression evidence: pre-place an unrelated committed segment at the derived ID and prove the old mixed segment, cursor, and live frame remain untouched; cover two acknowledged segments with identical survivors but different config/exporter policy; assert no old-row delete or cursor repoint occurs unless the existing candidate is the exact intended successor.

[P1] Mandatory reconciliation regresses a delivered checkpoint to pending

Action: Change

Gate/contract: Plan §§3.2, 3.4, 4.3, and 4.7; ADR 0004; G03 failure injection. The new segment must inherit per-exporter delivery state, restart must preserve checkpoints, and replay must remain logically idempotent.

Evidence: the normal build copies exporter states at `src/milhouse/spooling/compaction.py:214-247`, but after a crash between new-file publication and the ledger swap, mandatory reconciliation registers the orphan with every exporter `pending` (`src/milhouse/spooling/reconcile.py:2165-2185`). The retry path at `src/milhouse/spooling/compaction.py:347-371` accepts that row and never restores the old row's states. On this exact integrated tree I marked the old segment `delivered`, forced the swap transaction to fail after publication, ran real `DurableSpool` reconciliation, and retried compaction. The state sequence was `delivered -> [old delivered, orphan pending] -> new pending`. The added crash test at `tests/unit/test_spool_compaction.py:549-583` skips mandatory reconciliation and does not assert exporter state, so it misses the documented crash path.

Impact: a terminal destination confirmation is forgotten and already delivered records become eligible for duplicate external delivery after restart, violating checkpoint/replay recovery semantics.

Exact correction: when and only when a verified reconciled orphan is the exact intended successor, transactionally restore the old segment's exporter statuses and required immutable metadata before deleting the old row; refuse other existing rows. Preserve terminal delivery state through the same conditional ledger swap.

Required regression evidence: deliver a mixed segment, crash after compacted-file publication/before swap, run mandatory writer reconciliation, retry, and prove exporter states, cursor binding, file identity, and logical delivery count are unchanged. Exercise multiple exporters with mixed `pending`/`delivered`/`failed` states.

[P1] The new compaction lineage persists public unsalted identifier hashes

Action: Change

Gate/contract: Plan §§4.2, 4.4, and 4.7; ADR 0007; SECURITY privacy invariants. Persisted identifier derivatives must use an installation-keyed pseudonym, never a public unsalted hash.

Evidence: `src/milhouse/state/audit.py:71-81` explicitly computes plain `SHA-256(value)`, even while acknowledging that the accepted ID grammar admits secret-shaped values. The new `record_compaction` path stores both old and new values through that helper at `src/milhouse/state/audit.py:173-223`, and `tests/unit/test_spool_compaction.py:422-438` locks the public hashes into expected behavior. A low-entropy or dictionary-known ID is recoverable offline and stable across installations.

Impact: compaction adds two reversible-by-dictionary, cross-installation-correlatable identifier derivatives per rewrite to SQLite, expanding the base audit-privacy defect rather than meeting the keyed-pseudonym contract.

Exact correction: store an installation-keyed HMAC pseudonym (with the required key/epoch authority) or no identifier derivative at all; define safe treatment for already persisted unsalted rows. Do not call a public digest a privacy fingerprint.

Required regression evidence: use low-entropy and secret-shaped synthetic IDs to prove neither raw IDs nor public SHA-256 values are stored; prove different installations produce different pseudonyms, authorized same-installation correlation still works, and upgrade/legacy audit-row handling does not retain the vulnerable representation.

### Evidence run

- Live GitHub state immediately before posting: PR open; head/base unchanged; zero issue comments; zero formal reviews; one commit and its tip equals the reviewed full head SHA.
- Exact-head hosted checks: all 20 completed successfully, including `required-ci`, `test`, DCO, CodeQL, gitleaks, dependency review, packaging, quality, compatibility, and artifact smoke. No legacy status contexts exist.
- Commit DCO: the exact head contains a `Signed-off-by` trailer.
- Local exact integrated tree: `git diff --check` passed; 127 targeted compaction/reconciliation tests passed under Python 3.13.2. The canonical local wrapper could not run because the available `uv` is 0.7.10 while the repository requires 0.11.29, so I do not claim an independent local rerun of the PR's full 4,628-test/coverage totals.
- Independent adversarial reproductions above both completed against integrated commit `1f307b892bcb8bd2d03b133cf9ff8055b134a0c8`; the green suite does not cover them.

### Missing / externally pending gate evidence

- G03 remains in progress. This head cannot advance slice 5c while the three P1s remain.
- A fresh exact-full-SHA independent review is required after remediation. Physical supported-host crash evidence and the remaining slice-7 failure/concurrency evidence remain outside this PR.

### Residual risk

I did not repeat the entire suite locally because the locked `uv 0.11.29` runtime is unavailable in this environment; exact-head hosted gates are green, and the targeted/reproduction evidence is sufficient for the FAIL verdict. No source, status, review approval, merge, or publication state was changed by this review.

**Verdict: FAIL**

PR #74, opened 2026-08-01
docs(g03): specify the supported-host physical-crash test procedure

that1guy15:
Documents the manual, supported-host physical-crash tests that remain pending for G03 acceptance (owner-provided, E06) — the "circle back later" artifact.

`docs/gate-evidence/G03-physical-crash-procedure.md` specifies **nine test cases** (kill before rename, kill after commit, ledger-transaction kill, cursor/checkpoint kill, exporter-checkpoint kill, concurrent-writers-under-kill, compaction kill at three sub-boundaries, retention-prune kill, torn-tail power-loss), each with setup, the exact kill point, restart+reconcile verification, and pass criteria — and maps each to the deterministic in-process harness test that already covers the *logic* (so this procedure adds the real-`SIGKILL`/power-loss evidence the harness can't reproduce). Includes the environment to record per run and when each case is runnable now (library scripts) vs. needs a real host.

Docs only; linked from G03.md and the gate-evidence README; `docs-check` green. Not self-merged.

### DIFF
--- a/docs/gate-evidence/G03-physical-crash-procedure.md
+++ b/docs/gate-evidence/G03-physical-crash-procedure.md
@@ -1,0 +1,165 @@
+# G03 supported-host physical-crash test procedure
+
+Status: **procedure only — not yet executed.** This document specifies the manual, supported-host
+physical-crash tests that remain pending for G03 acceptance (owner-provided evidence, E06). It is a
+companion to [G03.md](G03.md); run it when a supported clean host is available (before G03 acceptance,
+and again at the RC soak).
+
+## Why this exists
+
+The automated [crash/concurrency/failure-injection harness](../../tests/unit/_g03_harness.py) proves
+every durable-write boundary **deterministically** by in-process fault injection and
+restart-by-reconciliation: a fault is raised at a chosen syscall or SQLite statement, and a fresh
+writer's mandatory reconciliation is treated as the "restart". That proves the *logic* is correct.
+
+It does **not** prove behaviour under a **real** process death — a `SIGKILL` or a host power-loss
+that stops the process at an arbitrary machine-code instruction, with in-flight page-cache and
+filesystem-journal state that no in-process test can reproduce. G03 requires that "a kill before
+rename exposes no partial batch" and "a kill after commit loses no acknowledged record" hold against
+those real kills on each supported filesystem. This procedure produces that evidence.
+
+The two are complementary: the harness is the fast, always-on regression; this procedure is the
+periodic, host-specific acceptance evidence.
+
+## Environment to record (per run)
+
+Capture all of the following into the run's evidence file, because crash behaviour is
+filesystem- and platform-specific:
+
+- OS name + exact patch/build (e.g. `Ubuntu 24.04.1`, `macOS 14.6.1 (23G93)`), architecture.
+- Filesystem and mount options of the directory holding `STATE_ROOT` (e.g. `ext4`, `apfs`,
+  `data=ordered`, `barrier=1`). Power-loss results are only meaningful with write barriers enabled.
+- Python version, and the exact repository commit under test (`git rev-parse HEAD`).
+- For power-loss cases: the exact method used to cut power (hardware switch, hypervisor
+  `stop`/`reset`, `echo b > /proc/sysrq-trigger`), since a clean hypervisor stop that flushes caches
+  is a weaker test than a true power cut.
+
+## Method
+
+Each case uses three roles:
+
+1. **Setup** — create a fresh `STATE_ROOT` (an initialized control database, `commit.lock`, and an
+   empty `spool/` tree) and any preconditioning segments, using the W03 library API exactly as
+   `tests/unit/_g03_harness.py::build_spool` and `commit_segment` do. Until the `milhouse` CLI exists
+   (W06), the driver is a short Python script that imports `milhouse.spooling` / `milhouse.state`.
+2. **Kill** — run a **driver subprocess** that performs one durable-write operation and, at the chosen
+   boundary, either (a) is sent `SIGKILL` by a controller, or (b) is running when the host loses
+   power. The boundary is reached by having the driver signal readiness (write a sentinel byte to a
+   pipe / touch a file) immediately before the target syscall, then block; the controller kills it on
+   the sentinel. This makes the kill point exact and repeatable without the driver ever completing the
+   operation.
+3. **Restart + verify** — from a new process, open the same `STATE_ROOT` (constructing a
+   `DurableSpool` runs mandatory reconciliation — the real restart path) and assert the case's pass
+   criteria against the ledger rows, the on-disk spool, and a `replay_segments` pass.
+
+Record, for every case: the pre-kill state, the post-kill on-disk artifacts (before restart), the
+post-restart state, and PASS/FAIL. Run each case **≥5 times** per filesystem; a single pass is not
+sufficient because the kill can land in slightly different sub-states.
+
+## Test cases
+
+Each maps to a G03 assertion and to the deterministic harness test that already covers the logic.
+
+### C1 — Kill before the atomic rename exposes no partial batch
+- Harness analogue: `test_g03_failure_injection.py::test_a_kill_before_rename_exposes_no_partial_batch`.
+- Setup: empty spool.
+- Kill: driver stages+fsyncs the segment temporary, signals readiness, blocks *before* the atomic
+  rename; controller `SIGKILL`s it (or power is cut here).
+- Restart + verify: **no committed `pending/<day>/<batch>.jsonl` exists**, **no `_segments` row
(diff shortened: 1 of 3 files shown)


that1guy15:

**Verdict: FAIL**

- Reviewed head: **a2e71b7d15e03eb22d4b44b0cf7c2e99b76ea457**
- Reviewed base: **b2822f3cb8b2cc99283778840b6c8a10d79af8e9**
- Integrated-tree basis: the reviewed head has the reviewed base as its sole parent; all three changed files were inspected.
- Scope: W03/G03 crash, recovery, privacy-expiry, audit, evidence truth, and supported-host procedure completeness.

### Findings

#### [P1] Retention crash recovery re-adopts privacy-expired bytes indefinitely
**Action: Change**

**Gate/contract:** Plan sections 4.3 and 4.7 make record-class privacy expiry a hard upper bound; ADR 0007 says spool and projections honor immutable expiry; the new procedure’s global criteria say expired data is never retained past its deadline and convergence completes through restart/reconcile.

**Evidence:** docs/gate-evidence/G03-physical-crash-procedure.md:129-153 instructs C8 to kill after the row-first ledger deletion, accept reconciliation re-registering the expired orphan, and rely on a later retention pass. On the exact head, _prune_segment commits the row deletion before unlink, while DurableSpool reconciliation blindly re-registers the remaining valid file. A synthetic exact-tree reproduction observed:

    observation_after_deadline 2026-07-28 12:00:04+00:00
    post_crash [] True
    restart_states [(1, ['batch-a'], True), (2, ['batch-a'], True), (3, ['batch-a'], True)]

The checked-in test_an_interrupted_unlink_leaves_a_re_registerable_orphan_that_reconverges also passes only because it manually invokes a separate later retention_apply; restart/reconcile alone never converges.

**Impact:** A process or host crash in the documented C8 window restores privacy-expired payload bytes to the live ledger and can retain them without a bound. This contradicts the hard privacy deadline and makes the proposed procedure internally impossible to pass as written.

**Exact correction:** Persist a privacy-safe durable prune/tombstone intent before removing the row, and make reconciliation finish the authorized unlink (or otherwise prove a bounded recovery protocol that never re-registers an expired retired segment). Do not depend on an eventual scheduler pass.

**Required regression evidence:** A real-process kill after the prune transaction and before unlink, followed only by repeated supported restart/reconcile cycles, must remove the expired file within a specified bound while preserving cursor position/revision and audit truth.

#### [P1] The bounded successor probe still strands mixed-expiry data
**Action: Change**

**Gate/contract:** Plan sections 4.3 and 4.7 require mixed-expiry compaction to enforce each frame’s privacy deadline; G03 cannot pass with privacy-expired frames stranded.

**Evidence:** C7 at docs/gate-evidence/G03-physical-crash-procedure.md:117-127 tests at most the primary derived-ID collision. The integrated code uses 1,024 caller-predictable candidates (src/milhouse/spooling/compaction.py:78-90 and 377-448) and returns collision_bound with the old mixed segment intact when all are occupied. The checked-in adversarial test test_compaction_reports_a_collision_when_every_candidate_is_foreign passes while explicitly asserting that the old mixed segment remains. Reporting the condition does not remove the expired frames or impose a retry bound.

**Impact:** A feasible prepopulation of deterministic batch IDs permanently prevents compaction and retains expired frames past the privacy ceiling. The procedure can pass its one-collision case while missing the actual hard-bound failure.

**Exact correction:** Use a successor allocation/retirement protocol whose namespace cannot be exhausted by caller-controlled committed IDs, while retaining exact-content verification and restart idempotency.

**Required regression evidence:** Exercise the real production bound (not a reduced monkeypatch), occupy every previously predicted candidate with foreign committed/orphan segments, and prove the expired frames are removed without touching foreign data or losing/duplicating live records, including after kills at all C7 sub-boundaries.

#### [P1] Audited compaction still has no attributable production lineage
**Action: Change**

**Gate/contract:** Plan section 4.7 requires compaction to record old/new hashes and lineage transactionally, and persisted identifier derivatives must be installation-keyed HMAC pseudonyms.

**Evidence:** C7 (docs/gate-evidence/G03-physical-crash-procedure.md:117-127) calls the operation audited but never verifies the audit rows or their old-to-new attribution. Production _swap_ledger calls record_compaction without a pseudonymizer, so both resource values remain NULL; docs/gate-evidence/G03.md:64-73 admits this. Deferring the key to W06 does not resolve the gate: W06 depends on G05, while G05 depends on G03, and no accepted amendment changes that dependency.

**Impact:** A crash test can report C7 PASS without proving which segment was compacted into which successor. G03 remains blocked by a downstream dependency cycle even if every physical-host case passes.

**Exact correction:** Wire a trusted installation-keyed pseudonym source into production W03 compaction before G03, or obtain an owner-approved plan/dependency amendment that preserves attributable audit semantics and breaks the cycle.

**Required regression evidence:** For every C7 kill boundary, verify exactly two non-null canonical keyed lineage resources survive or roll back atomically with the swap, can be attributed to the intended old/new IDs using the authorized key, and contain no raw/public-hash identifier.

#### [P2] The physical-kill procedure does not supply reproducible internal kill controls or a fixed host matrix
**Action: Verify**

**Gate/contract:** Gate evidence must be current and reproducible; supported-platform evidence must record the exact environment and must not turn placeholders into passing evidence.

**Evidence:** The generic method at docs/gate-evidence/G03-physical-crash-procedure.md:41-57 says a driver signals immediately before an internal syscall and blocks, but no controller/driver is supplied for the SQLite commit, cursor/checkpoint transaction, post-export/pre-CAS window, compaction swap, or retention window. Those points are inside public W03 calls; killing before the call is not killing during/after it, and C5’s side-file gives a race rather than the claimed exact post-confirmation/pre-CAS stop. C9 at lines 138-144 also does not say how the production temp-write/rename path yields a torn committed file instead of merely a staged temporary. Finally, lines 155-163 change “each supported filesystem” into “every tested supported filesystem” without enumerating the required OS/filesystem matrix, and cite E06 even though the current ledger scopes E06 to G06/G15-G18 rather than G03.

**Impact:** Independent operators can execute materially different experiments and still call them the same evidence; some listed cases cannot reach the claimed production boundary from the instructions alone.

**Exact correction:** Check in a bounded controller/driver with immutable candidate hash, explicit synchronization/interposition for each boundary, pre/post state assertions, timeout/failure behavior, and result schema. Define the minimum macOS/Ubuntu/filesystem matrix and correct the E06/G03 authority mapping. Separate true power-cut evidence from SIGKILL.

**Required regression evidence:** Run the checked-in driver from the exact installed candidate on every declared host/filesystem, retain environment/artifact hashes and per-repetition outcomes, and have an independent reviewer reproduce at least one run.

### Test and check evidence

- git diff --check: passed.
- Direct internal link validation: 71 files / 244 links passed.
- Focused exact-tree tests: 8 passed across G03 failure injection, retention interruption, and compaction crash paths.
- Probe-exhaustion adversarial test: 1 passed, reproducing collision_bound with the old segment retained.
- Canonical local ./scripts/run_make.py docs-check: not claimable on this host because the repository locks uv 0.11.29 and the resolved local executable is uv 0.7.10. The direct link checker passed instead.
- GitHub at review time: DCO and 17 other checks succeeded; test remained in progress, so strict required-ci had not completed.
- Commit contains an author-matching Signed-off-by trailer.

### Missing / externally pending evidence

- No supported-host SIGKILL or true power-loss run exists yet.
- No fixed driver/controller artifact or immutable per-host result packet exists.
- The production keyed compaction lineage remains absent.
- Required CI was still pending at review time.

### Residual risk

The existing deterministic tests are useful regression coverage but deliberately cannot establish kernel/filesystem durability under process death or power loss. Green CI does not resolve the reproduced privacy-expiry paths or the missing production audit attribution.

**Final verdict: FAIL.** Do not accept G03 from this head.

that1guy15:
Superseded by #167, merged to `main` as `b894100`.

This PR's unique content — `docs/gate-evidence/G03-physical-crash-procedure.md`, which existed nowhere else — was salvaged and brought current rather than dropped:

- **Reframed** from "pending for G03 acceptance" to the deferred **E06** (#108) specification. G03 was accepted 2026-08-07; supported-host physical-crash evidence moved to a later hardening lane and is not a gate blocker.
- **Removed case C7 ("kill during audited compaction")** — amendment A09 deleted `src/milhouse/spooling/compaction.py` and `tests/unit/test_spool_compaction.py`, so the case pointed at code and a harness analogue that no longer exist. C8/C9 renumbered; a scope note records the removal.
- **Corrected the driver guidance** — "until the `milhouse` CLI exists (W06)" is obsolete; the CLI now handles setup and verification, while the kill step stays at the library level because these boundaries sit inside a single durable operation.
- Every surviving harness reference was verified to still exist at that commit.

The remainder of this PR's diff was deliberately **not** taken: it predates G03 acceptance, so its `G03.md` edit would have deleted the accepted status record — the reason it went `CONFLICTING`.

Timing note: this became worth landing now because dedicated hardware for the project exists, which is the one thing E06 was waiting on.

Issue #123, opened 2026-08-08
W04 G04b: idempotent ClickHouse delivery ledger (ClickHouseExporter + ledger-gated export)


that1guy15:
## W04 G04b — first increment: idempotent ClickHouse delivery ledger

Now dependency-unblocked (G03 accepted 2026-08-07). Part of #93.

### Problem
The spool owns a complete, G03-certified exactly-once delivery machine — `deliver_segment` (fenced CAS `pending/failed→delivered`), `replay_segments` (drain), the `_segment_exporters` ledger — keyed on exporter id `"clickhouse"`. The ClickHouse side has a **separate, ungated** bulk `export_records` that re-scans and re-inserts every committed segment each run, delegating idempotency to `records`' `ReplacingMergeTree`. Because `feedback_transitions` is append-only, every re-export writes **duplicate transition rows** (`storage/exporter.py` docstring flags this). No `ClickHouseExporter` implements the `Exporter` protocol, so the ledger/drain has nothing to drive for ClickHouse.

### Scope (this increment — fully offline-testable, in Required CI)
- Add `ClickHouseExporter(Exporter)`: `exporter_id == "clickhouse"`, `deliver(record, frames)` reuses the existing row-mappers + `client.insert`, raises `StorageError` on failure.
- Repoint `storage export` onto `replay_segments(..., exporters={"clickhouse": …}, delivery_status="pending", now=…)` so export is **ledger-gated** and re-export of a delivered segment is a no-op CAS (eliminates the `feedback_transitions` duplicate-row cost).
- Tests via `FakeClickHouseClient` + the `test_spool_exporter`/`test_spool_replay` harness: first drain delivers + inserts once; second drain → `already_delivered` with zero new inserts; a raising exporter → `failed`, healthy re-drain (clock advanced ~24h) → delivered exactly once.

### Deliberately NOT in this increment (owner/host-gated)
- **24-hour outage drain (live)** — offline kernel here; live confirmation needs a running ClickHouse + the opt-in `tests/live` smoke.
- **Native backup/restore** verification (record IDs + exporter checkpoints + migration state) — ClickHouse `BACKUP`/`RESTORE` are server operations → **E06** clean-host evidence (#108). Feedback-state backup parity is plan-deferred to G16.
- Gate acceptance (G04b Passed) — requires the live evidence above + owner acceptance.

### Reuses (no reinvention)
[...]

Issue #108, opened 2026-08-03
E06 — Supported-host physical-crash evidence (DEFERRED)


that1guy15:
Owner/host-provided power-loss / SIGKILL crash evidence on supported macOS/Ubuntu hosts. **Deferred to a hardening lane per the owner's 2026-08-03 direction**; the in-process crash/fault-injection harness already proves the boundaries deterministically. Not a blocker for the re-scoped G03.

---
*Work package:* `E06` · *Gate:* G06/G15–G18 · *Depends on:* W03 core · *State:* Deferred
*Authoritative source:* [implementation-plan.md](docs/implementation-plan.md) / [implementation-status.md](docs/implementation-status.md)

that1guy15:
## E06 — consolidated live / host-gated validation checklist

Registering the full set of **host-gated evidence** to run in one **overall live-testing pass** (owner's 2026-08-11 direction). This issue is the single tracking point spanning (a) the original E06 supported-host physical-crash evidence and (b) the W04 **G04a/G04b** ClickHouse live exit-criteria that Required CI cannot prove offline. **None of these are claimed passed yet; the relevant gates stay `In progress` / not-passed until this pass is run and recorded.**

### How to run the automated live smokes
Opt-in, self-skip in Required CI. Set `MILHOUSE_LIVE_CLICKHOUSE=1` (+ `MILHOUSE_CLICKHOUSE_*` creds) against the loopback reference deployment (`ops/clickhouse/` compose, which now also provisions the `backups` disk + durable `clickhouse-backups` volume), then run `tests/live/test_clickhouse_smoke.py`.

### A. Automated live smokes (already written; just need a real ClickHouse)
- [ ] **G04a — fresh migrate + idempotency + checksum enforcement** — `test_live_fresh_migrate_status_idempotent_and_checksum_enforcement`
- [ ] **G04a — default/empty-password account denied** — `test_live_default_empty_password_account_is_denied`
- [ ] **G04b — export round-trips records + derives feedback state** — `test_live_export_round_trips_records_and_derives_feedback_state`
- [ ] **G04b — native BACKUP → DROP → RESTORE round-trips + reconciles** (#126) — `test_live_native_backup_restore_round_trips_and_reconciles`
- [ ] **G04b — installation ownership claims + rejects a cross-installation** (#128) — `test_live_installation_ownership_claims_and_rejects_a_cross_installation`

### B. Procedural host-gated validations NOT covered by the smokes (run manually during the pass)
[...]

Issue #93, opened 2026-08-03
W04 — Secure ClickHouse, migrations, repository, and recovery


that1guy15:
Loopback Compose, packaged SQL migrations + checksum runner, typed authenticated client/exporter/repository, backup/restore.

**G04a** (auth/migrations/checksums) depends only on G02 — **dependency-ready to start now**. **G04b** (idempotent delivery, 24h outage drain, native restore) needs G03.

---
*Work package:* `W04` · *Gate:* G04a / G04b · *Depends on:* G02 (G04b also needs G03) · *State:* Ready
*Authoritative source:* [implementation-plan.md](docs/implementation-plan.md) / [implementation-status.md](docs/implementation-status.md)

PR #124, opened 2026-08-08
feat(storage): W04 G04b idempotent ClickHouse delivery ledger (ClickHouseExporter)

that1guy15:
## What & why

First increment of **W04 G04b** (#123): an idempotent, outage-recovering ClickHouse delivery ledger — now dependency-unblocked by G03.

The spool already owns a complete, **G03-certified** exactly-once delivery machine (`deliver_segment` fenced CAS, `replay_segments` drain, `_segment_exporters` ledger) keyed on exporter id `"clickhouse"`, but nothing implemented the `Exporter` protocol for ClickHouse and `storage export` **bypassed the ledger** — re-scanning and re-inserting every committed segment each run, so the append-only `feedback_transitions` table accumulated a duplicate row per re-export. This wires the two together.

## Changes

- **`storage/delivery.py`** (new) — `ClickHouseExporter` satisfies the spool `Exporter` protocol; `deliver()` reuses `export_records` (per-record `LOCAL_CLICKHOUSE` egress + native-column insert) and raises on failure so `deliver_segment` records a retryable `failed`.
- **`cli/root.py`** — `storage export` now drives `replay_segments(delivery_status=None)` through the exporter: ledger-gated, exactly-once-logical, retries pending+failed, idempotent no-op for delivered. Per-outcome output, exit 1 on any `failed`, informational stderr note on `expired`.
- **`cli/views.py`** — removed the now-dead `read_trusted_records`/`TrustedRecordScan`.
- **`storage/exporter.py`** — docstring truth fix (the ledger now owns per-segment de-dup).
- **Tests** — `test_storage_delivery.py` + rewritten `test_cli_storage.py` on the real spool machine + a fake client: deliver-once, re-drain-no-new-inserts, fail→recover-exactly-once, and the expired-withheld note.

## Why `delivery_status=None` (not `"pending"`)

A pending-only drain silently skips a segment whose delivery went **`failed`** during an outage (the ledger filter is exact-match) — breaking recovery. `None` drains all committed segments; `deliver_segment`'s CAS makes already-delivered ones idempotent no-ops and retries pending+failed — the proven G03 double-replay recovery shape. (Bounded by retention pruning delivered segments; a future "not-delivered" ledger filter could optimize the re-reads — noted in-code.)

## Evidence

- `make test`: **4808 passed / 4 skipped**.
- `make quality` + `make secret-scan`: green.
- **Independent gate-review** (report-only): verdict `externally_pending`, **no actionable P0/P1/P2**. Confirmed idempotency, outage-recovery, a real production caller (no D14 uncalled-subsystem trap), and fail-closed on tamper/expiry/restricted — this path *adds* the hard privacy-expiry withholding the old bulk path lacked. (Two P3 "Consider" nits raised were fixed here.)

## NOT claimed (owner/host-gated)

The **live 24-hour outage drain** and **native backup/restore** (G04b Obligations 2-live and 3) require a running ClickHouse + **E06** clean-host evidence (#108). This increment covers only the offline idempotent-delivery-ledger portion. **G04b remains In progress / not passed** — this PR does not claim a gate pass and needs your review before merge.

Refs #123
Refs #93


that1guy15:

Exact head SHA: `aa4761700c0f1365762997c95a704ecd1d6b0ae1`

Reviewed base SHA: `15c78810549e505b7441a3d7a501fd264634e963`

Integrated candidate: GitHub merge ref `543f8c650977774762bd3590c172f29903851566` has parents `15c78810549e505b7441a3d7a501fd264634e963` and `aa4761700c0f1365762997c95a704ecd1d6b0ae1`; its tree is byte-identical to the head tree.

### Findings

[P1] The concrete exporter is not idempotent at the delivery protocol's required crash and concurrency boundaries

Action: Change

Gate/contract: W04/G04b requires repeated export/replay to be logically idempotent. The already-accepted delivery protocol also states that a crash after destination confirmation leaves the row retryable and therefore exporters **must** be idempotent (`src/milhouse/spooling/exporter.py:28-36`).

Evidence: `ClickHouseExporter.deliver()` delegates directly to `export_records()` (`src/milhouse/storage/delivery.py:59-69`), while that function explicitly says it is not independently idempotent and inserts deterministic feedback transitions into the raw append-only `MergeTree` (`src/milhouse/storage/exporter.py:12-22,145-175`). The ledger CAS happens only after those inserts. Two review-only adversarial reproductions against the exact integrated tree both produced two physical `feedback_transitions` inserts for the same deterministic transition: (1) two simultaneous drains both passed the shared-barrier reload and reached the destination before one CAS won; (2) a simulated process failure after destination confirmation but before `_advance_status_locked`, followed by recovery replay. The checked-in tests cover only a sequential second run after the first checkpoint has already committed, so they cannot detect either required failure boundary.

Impact: the claimed idempotent ClickHouse exporter and duplicate-transition elimination do not hold under the very at-least-once recovery window the protocol documents. Append-only history, counts, storage, and a native backup can contain duplicate logical transitions even though `feedback_current` happens to derive the same latest state. This does not discharge the offline idempotent-delivery part of G04b.

Exact correction: make every ClickHouse projection retry-idempotent by deterministic record/item/transition identity before treating destination return as confirmation. The solution must cover multi-table partial failure, concurrent drainers, and confirmation-to-ledger-checkpoint crash; the SQLite CAS alone cannot close a remote-destination crash window. Add a forward migration or retry-stable destination deduplication contract rather than changing an applied migration in place.

Required regression evidence: run two independent control connections/drainers against one destination and inject failure after the final destination insert but before the ledger CAS. After retry, prove one logical record, one item, and one transition per deterministic ID through every supported history/current query surface, plus the correct terminal exporter checkpoint. Repeat against a real supported ClickHouse.

[P1] Remove the committed Claude session-history locator from the public commit

Action: Change

Gate/contract: the repository safety boundary forbids committing session histories and private identifiers (`AGENTS.md:98-103`), and the provenance policy excludes raw Claude/session material from the public tree and history.

Evidence: the sole commit at this head includes `Claude-Session: https://claude.ai/code/session_...` in public commit metadata. Hosted gitleaks and DCO are green, but those checks do not make a committed session locator permissible; I did not follow or inspect the external session.

Impact: the public Git history permanently publishes a private agent-session identifier and a locator that may expose or later be made to expose session content. This is a direct locked privacy/provenance boundary violation.

Exact correction: invalidate/revoke any shared session locator if applicable, rewrite the PR commit to remove the `Claude-Session` trailer, and ensure the replacement commit retains a valid DCO sign-off without copying session metadata elsewhere.

Required regression evidence: scan every commit introduced by the replacement head, not only tree files, for agent-session URLs/identifiers; show zero matches and a passing DCO check.

[P2] `storage export` can report success while exporting none of a valid committed segment

Action: Change

Gate/contract: the command says it drains every committed segment, and ClickHouse is a rebuildable analytical copy of retained spool data. A successful export must not silently omit acknowledged records.

Evidence: `replay_segments(delivery_status=None)` lists every segment, but `deliver_segment()` attempts only exporter rows already present in that segment (`src/milhouse/spooling/exporter.py:280-323`). The CLI then counts only `delivered`, `already_delivered`, `failed`, and `expired`, and exits nonzero only for `failed` (`src/milhouse/cli/root.py:550-613`). A review-only reproduction committed a fully valid one-record segment with `required_exporters=()`—an accepted spool header shape—then ran the exact replay/export mapping: the report said one segment/one record, returned zero delivery attempts, performed zero ClickHouse inserts, and would leave the CLI at exit 0.

Impact: spool-only or otherwise valid retained segments without a pre-existing `clickhouse` ledger row can be omitted from a rebuild/backfill while the operator receives a successful record count. This is a bounded but material compatibility and recovery defect in the replacement for the previous bulk export path.

Exact correction: define and implement an authoritative ClickHouse backfill contract for eligible retained segments without a `clickhouse` row, compatible with the header's immutable delivery-policy rule. If this command is intentionally limited to pre-registered ClickHouse obligations, reject or explicitly report every unhandled segment and exit nonzero instead of counting it as a successful drain; provide the separately specified rebuild path before claiming spool recoverability.

Required regression evidence: cover valid segments with zero exporters, only a different exporter, and `clickhouse` plus another exporter; assert exact ClickHouse rows, per-exporter outcomes, operator output, and exit status. Include a spool-only-to-full recovery/rebuild scenario.

### Test and check evidence

- Exact-tree local run with pinned `uv 0.11.29` / CPython 3.14.6: candidate's 4,808 checked-in tests plus 3 review-only adversarial tests all passed; 4 live tests skipped. The three adversarial tests intentionally assert the reproduced defective behavior described above.
- Changed-area delivery/CLI/exporter/replay suites: 55 passed.
- `docs-check`: 72 files, 301 links, and 80 external probes passed.
- `skill-check`: five canonical skills and discovery aliases valid.
- `git diff --check`: passed.
- Live hosted head checks: all 20 terminal; Required CI, test, quality, compatibility, package, audit, gitleaks, dependency review, DCO, and CodeQL are successful/accepted.
- Commit tip pagination matches the exact head; DCO is green. The commit is unsigned, which is not itself a DCO failure.

### Missing or externally pending gate evidence

- The PR correctly leaves G04b In progress. The real supported-ClickHouse 24-hour outage drain and native backup/restore evidence remain externally pending and are not counted as defects in this offline increment.
- I did not run a live ClickHouse server locally. Real-server retry/deduplication behavior remains unverified and is required after the P1 correction.

### Residual risk

The current-state view masks duplicate transition rows, so ordinary happy-path queries and green CI do not expose the recovery defect. Multi-table partial confirmation and future history/backup consumers remain unsafe until destination-level idempotency is proven.

VERDICT: FAIL

that1guy15:

Exact head SHA: **54163cb9fe48f99c0597f9e17d74a3b2ab1ec671**

Reviewed base SHA: **15c78810549e505b7441a3d7a501fd264634e963**

Integrated candidate: GitHub merge ref **48b96697762372e5cf7158f43529abf86606f41a** has parents **15c78810549e505b7441a3d7a501fd264634e963** and **54163cb9fe48f99c0597f9e17d74a3b2ab1ec671**; its tree **08deed929921f62a918c57b2723a2749821e1762** is byte-identical to the head tree.

### Findings

[P1] Migration 0005 can drop a transition exported during its copy-and-swap window

Action: Change

Gate/contract: Plan section 4.5 and ADR 0005 require migrations not to silently drop data, while G04b requires outage/replay recovery to converge without lost logical history.

Evidence: migration 0005 takes a snapshot-style INSERT SELECT from the live feedback_transitions table into a scratch table, then EXCHANGEs the two names and drops the old table (src/milhouse/storage/migrations/0005_feedback_transition_dedup.sql:28-34). The storage migrate command takes no global migration/export fence (src/milhouse/cli/root.py:487-501), while storage export can insert through a separate process. A transition inserted into the original table after INSERT SELECT completes but before EXCHANGE remains only in the old table; after the swap that table is named feedback_transitions_0005 and the final DROP deletes the transition. A review-only schedule over the exact packaged SQL reproduced final_live=[existing-transition], injected=true, lost=true. The live smoke starts from a fresh empty database and therefore does not exercise an upgrade with a concurrent writer (tests/live/test_clickhouse_smoke.py:123-170).

Impact: an acknowledged feedback transition can disappear from the ClickHouse history/current projection during a normal schema upgrade even though its SQLite exporter row may already be terminal delivered. Ordinary replay then skips that segment, so the analytical store does not self-heal without manual ledger repair or a full rebuild. This violates the locked no-data-drop migration contract and blocks this G04b increment.

Exact correction: fence every ClickHouse exporter across the full copy, swap, verification, and old-table retirement, using maintenance authority shared by every production export path; or use a ClickHouse-native online evolution that dual-writes/catches up all deltas before cutover. Do not drop the old table until the new table is proven to contain the complete pre-cutover history, and make interruption/retry safe at every statement boundary.

Required regression evidence: with two independent clients, pause migration after the copy, insert and confirm a transition through the real exporter, then resume the swap. Prove that the transition exists exactly once logically, feedback_current is correct, the exporter checkpoint remains consistent, and retry after failure at each copy/swap/drop/ledger boundary preserves all rows. Run this on both supported ClickHouse lines.

[P2] A completely failed export reports records that never reached ClickHouse

Action: Change

Gate/contract: The CLI contract and this command require stable, truthful operator output; an incomplete delivery must not be represented as exported records.

Evidence: the record-count calculation includes every non-expired ClickHouse attempt, explicitly including failed attempts because they might reach ClickHouse on a future run (src/milhouse/cli/root.py:586-600), and emits that value as records (src/milhouse/cli/root.py:614-633). A review-only exact-head reproduction made the first ClickHouse insert raise: the client recorded zero physical inserts and the command exited 1, but stdout reported segments=1 records=1 delivered=0 failed=1. The checked-in failure test asserts failed=1 and zero inserts but does not assert the records field (tests/unit/test_cli_storage.py:226-244).

Impact: automation and operators cannot distinguish records confirmed present from records merely eligible for retry. The JSON contract can overstate warehouse contents during the exact outage and partial-failure cases this change is intended to make observable.

Exact correction: define records as confirmed-complete records and count only delivered/already-delivered segments, or rename the field to an unambiguous attempted/eligible count and add separate confirmed and retryable record counts. Do not infer future delivery as present delivery.

Required regression evidence: assert the full human and JSON payload for failure before the first insert and after each multi-table partial-insert boundary, then assert that a later successful retry moves the records into the confirmed count exactly once.

### Test and check evidence

- Exact integrated tree, pinned uv 0.11.29 / CPython 3.14.6: full suite **4,815 passed / 4 skipped**.
- Changed-area delivery, CLI, migration, schema, packaging, and live-smoke collection: **42 passed / 3 skipped**; the skips are the opt-in live ClickHouse cases.
- Review-only adversarial schedules reproduced both findings above.
- docs-check: **72 files, 301 links, 80 external probes passed** after retrying one transient connection failure.
- skill-check: five canonical skills and discovery aliases valid.
- secret-scan: private-identifier scan plus current-tree and 234-commit history scans passed; the prior Claude session-history locator is absent from this replacement commit.
- git diff --check passed; the integrated merge tree equals the exact head tree.
- Live hosted head checks: all **20** terminal and successful/accepted, including required-ci, test, quality, compatibility, package, audit, gitleaks, dependency review, DCO, and both CodeQL surfaces. Commit pagination tip matches the exact head. The commit is unsigned, which is not itself a DCO failure.

### Missing or externally pending gate evidence

- I did not run a live ClickHouse service locally. The migration copy/swap, physical ReplacingMergeTree collapse, and supported-version behavior remain unverified on a real server.
- The 24-hour outage drain and native backup/restore evidence remain externally pending as the PR states; G04b correctly remains In progress. Those deferred obligations are not counted as additional findings here.

### Residual risk

The deterministic destination keys appear to close the prior confirmation-to-checkpoint and concurrent-drain logical-duplicate defect, but the live physical-collapse claim remains host-gated. Multi-table partial confirmation and migration/export overlap remain unsafe until the P1 correction is proven on real supported ClickHouse versions.

VERDICT: FAIL

that1guy15:

Exact head SHA: **3ee23a71b88422a44c9dc5afd610dd84c5706778**

Reviewed base SHA: **15c78810549e505b7441a3d7a501fd264634e963**

Integrated candidate: GitHub merge ref **1b1a098494912c141d6bd8b11dddc5b74c6208ee** has parents **15c78810549e505b7441a3d7a501fd264634e963** and **3ee23a71b88422a44c9dc5afd610dd84c5706778**; its tree **cae01150e1f901ae369e9b7fe7182d266d2b08de** is byte-identical to the head tree.

### Findings

No actionable findings.

The replacement head closes the two findings from the preceding exact-head review:

- `storage migrate` now holds the installation's exclusive global commit barrier across the complete migration apply, while the sole production ClickHouse writer delivers beneath the shared side (`src/milhouse/cli/root.py:487-521`, `src/milhouse/cli/root.py:542-630`). The real-server adversarial runs described below verified both the live copy/swap fence and retry safety at every statement boundary of migration 0005.
- `storage export` counts only `delivered` and `already_delivered` batches as confirmed records; failed, expired, and obligation-free segments cannot inflate that count, and failed/unhandled work exits nonzero (`src/milhouse/cli/root.py:594-675`).

The earlier destination-idempotency correction also remains intact: retry-stable record/item/transition identities feed deduplicating ClickHouse engines; the transition table now uses `ReplacingMergeTree` keyed by `(item_id, transition_id)` (`src/milhouse/storage/migrations/0005_feedback_transition_dedup.sql:14-34`), and inserts explicitly disable asynchronous acknowledgement (`src/milhouse/storage/client.py:150-178`).

### Test and check evidence

- Exact integrated tree with pinned `uv 0.11.29` / CPython 3.14.6: **4,820 passed / 4 skipped**; the four skips are opt-in/live cases. The full suite includes the deterministic-ID crash-after-confirmation and concurrent-drain schedules, migration/export barrier tests, truthful CLI-output cases, schema/package checks, and prior W00-W03 coverage.
- Changed-area delivery, CLI, barrier, storage client/runner/schema, and packaging suites: **84 passed**.
- Live authenticated ClickHouse smoke on both supported lines — **26.3.17.56** and **25.8.28.1** from the repository's digest-pinned images: **3 passed per line**. This exercised fresh migration through version 5, checksum tamper refusal, authenticated/anonymous access behavior, record and feedback round trips, and physical transition collapse after re-delivery.
- Review-only live adversarial suite on both supported ClickHouse lines: **7 passed per line**. Six cases interrupted migration 0005 after each possible completed-statement boundary, inserted a newly confirmed transition after the simulated process death, retried, and proved both old and new transitions survived exactly once logically. The seventh paused a real migration after its copy, proved a real delivery remained blocked by the exclusive fence, resumed the swap/drop, and then proved one delivered ledger outcome and one transition row.
- `docs-check`: **72 files, 301 links, 80 external probes passed**. `skill-check`: five canonical skills and discovery aliases valid. Secret/private-identifier scans passed for **346 tracked text files**, the exact tree, and the **89-commit** fetched history. `git diff --check` passed and the exact review checkout remained clean.
- Live GitHub evidence: all **20** head checks are terminal and successful, including required-ci, test, quality, compatibility, package, audit, gitleaks, dependency review, DCO, and both CodeQL surfaces. Paginated commit collection ends at the exact head; every introduced commit has a DCO sign-off. The commit is unsigned, which is not itself a DCO failure.

### Missing or externally pending gate evidence

- This is an offline/incremental W04 correction, not gate acceptance. The 24-hour ClickHouse outage/drain and native backup/restore evidence remain owner/host-gated, and G04b must remain **In progress** until those separately tracked obligations pass.
- I did not exercise a real process kill between the remote insert acknowledgement and the SQLite ledger compare-and-set. The exact deterministic-ID retry schedule is covered offline, and each supported real server independently proved those duplicate IDs collapse, but the combined physical kill remains part of the longer host evidence rather than this review.

### Residual risk

The migration fence is installation-local and relies on `storage export` remaining the sole production ClickHouse row writer. Any future writer must acquire the shared side of the same barrier before this copy/swap migration shape remains safe. ReplacingMergeTree provides eventual physical collapse, so correctness-sensitive reads must continue using the repository's `FINAL`/deduplicated views. Neither residual is a new defect in this exact increment.

VERDICT: PASS WITH CONDITIONS

Condition: this head may satisfy only the reviewed offline increment; merging it must not mark G04b passed or waive the outstanding 24-hour outage and native backup/restore evidence.

PR #126, opened 2026-08-10
feat(storage): W04 G04b native ClickHouse backup/restore (offline core + CLI)

that1guy15:
## What & why
W04 G04b **native ClickHouse backup/restore** — the offline verification core + CLI. Advances the last G04b obligation (after the merged delivery-ledger increment #124). Closes #125.

## Changes
- **`storage/backup.py`** — `BackupSnapshot`, `snapshot_state` (raw record-id set + migration state), `verify_restored` (strict source↔restored equality — used by the round-trip test/live-smoke), `reconcile_delivered` (cross-store: every `clickhouse`-delivered segment's record ids, read from the surviving SQLite ledger + durable spool, are present in the restored store), and injection-safe `BACKUP`/`RESTORE`/`DROP` statement builders (fixed backup disk; only a bounded archive name interpolated).
- **`storage backup DEST`** / **`storage restore [--force] SOURCE`** CLI (storage group, exclusive-barrier-fenced per §10.3, require init). `restore` is an in-place DR restore: it **refuses to overwrite a non-empty analytical DB without `--force`** (§10.3 confirmation), drops+restores with it (an empty/absent DB restores cleanly without it), then **fails closed** unless the restored `_migrations` is checksum-consistent at the current defined head AND `reconcile_delivered` passes against the surviving ledger. Restore-to-a-new state root + the composite SQLite+manifest restore are **W16**.
- Tests: offline core (round-trip equality, fail-closed mismatch, reconciliation, builders, bad-identifier rejection) + a DR CLI suite driving the **actual** `storage restore` command + a host-gated live smoke case.

## Interpretation — please validate at gate acceptance
A ClickHouse native backup is ClickHouse-side only, so G04b's "matches **exporter checkpoints**" is scoped here to the **SQLite-ledger ↔ restored-ClickHouse reconciliation** (the manifest watermark + composite SQLite backup are W16); **feedback-state parity is deferred to G16**.

## Review trail
Built, then **two adversarial gate reviews** — the second found and drove a **P1** fix (the CLI `restore` verified against the *pre-restore* baseline, which would false-fail a real restore; redesigned to verify against the surviving ledger + migration head, with the `--force` overwrite guard). Gates green: `test` **4848 passed / 5 skipped**, `quality` + `secret-scan` clean. **Recommend a `/code-review ultra` on this PR as the final independent check** before merge.

## NOT claimed (owner/host-gated)
[...]

PR #128, opened 2026-08-10
feat(storage): W04 enforce one installation per ClickHouse (fail-closed ownership guard) (#127)

### LINKED ISSUE #127: W04/G04 P1: enforce single-installation ownership of a ClickHouse destination (cross-installation writer exclusion)
From the /code-review ultra on #126 (a P1 that predates it — affects the merged delivery increment #124 too).

**Gap:** the GlobalCommitBarrier is a LOCAL file lock (one host/state-root). It excludes writers WITHIN one installation, but does NOT coordinate two installations that point storage.export / storage.migrate / storage.backup / storage.restore at the SAME ClickHouse destination. Plan §§4.3/10.3 require writers excluded during backup/restore/migrate; a second installation could write/migrate/restore the shared destination concurrently → data loss / inconsistency.

**Fix options (owner architecture decision):**
1. Durably enforce ONE installation identity per ClickHouse destination — stamp the owning installation id in ClickHouse (e.g. a control row) and fail a second installation CLOSED. Backup/restore/migrate/export all share that authority.
2. OR destination-scoped coordination authoritative at the ClickHouse endpoint/database.
3. OR, if the supported model is one-ClickHouse-per-installation (local-first), DOCUMENT that as a hard supported-deployment constraint and add a fail-closed guard against a second installation.

**Decision needed:** is a ClickHouse destination shared across installations a supported scenario? If not, option 3 (document + fail-closed guard) is the bounded fix. This is a G04b acceptance consideration.

Refs #93, #123, #125.

that1guy15:
## What & why
Enforce **one installation per ClickHouse destination**, fail-closed — resolving **#127**, the cross-installation coordination gap #126 flagged. The local commit barrier serializes exactly ONE installation's writers; it is a machine-local lock and cannot fence a **second** installation (different state root, clone, or host) pointed at the **same** ClickHouse. Two such installs would each hold the single-writer lock while silently comingling and corrupting each other's data. This makes "one ClickHouse per installation" an **enforced invariant** rather than an unwritten assumption. (Owner chose this over shared-destination coordination.)

## Changes
- **`migrations/0006_installation_ownership.sql`** — a single-owner control table `{database}._installation`: one logical row at `id = 1` under `ReplacingMergeTree(claimed_at)` (newest claim supersedes on merge; a `FINAL` read sees the current owner pre-merge). `CREATE TABLE IF NOT EXISTS` (retry-safe). 0001–0005 byte-unchanged; migration set repins **5 → 6** across `schema.py`, the packaged manifest, and every pin test.
- **`storage/ownership.py`** — `read_owner` (owner, or `None` on unmigrated/absent/empty — existence-guarded like the runner's ledger read, reads through `FINAL`), `claim` (unclaimed→stamp; same id→no-op; **different id → fail-closed `MH_STORAGE_OWNERSHIP` unless `--reclaim`**), `require_owner` (different **or** unclaimed → fail-closed). Installation id validated as the bounded non-secret `mh_in1_` token before interpolation (structural literal, not parameter-bindable) — same discipline as the database identifier.
- **`cli/root.py`** — `storage migrate --reclaim` claims after applying migrations under the exclusive barrier, reporting `ownership: claimed|verified|reclaimed` (text + `--json`). `storage export`/`backup` call `require_owner` **before** writing. `storage restore` checks the restored owner **after** the schema gate and refuses a cross-installation restore with a **restore-specific `MH_STORAGE_RESTORE`** naming the manual `DROP` recovery — deliberately **not** the `--reclaim` guidance, which would *adopt* the foreign backup (empty-target-only restore ⇒ the operator drops the DB back to empty).
- **Docs** — migrate `--help` + `ops/clickhouse/README.md` document the reclaim/upgrade semantics and the foreign/failed-restore recovery.
- **Tests** — `test_storage_ownership.py` (claim/verify/reclaim; unclaimed + foreign `require_owner` fail closed; bad-id rejection; `FINAL` newest-claim read) + `test_cli_storage.py` ownership cases (migrate reports claimed/verified/reclaimed; export/backup refuse a foreign destination; restore refuses a foreign backup with the manual-`DROP` message, not `--reclaim`) + a host-gated live smoke case. `FakeClickHouseClient` models `_installation` so the owner travels through BACKUP/RESTORE/DROP.

## Review trail
Built, then an **independent adversarial gate review** — verdict **`externally_pending`**, **no P0/P1**, no fail-open or bypass reproduced (raises on unclaimed/mismatch/read-error, reads via `FINAL`, every writer takes it, takeover gated behind `--reclaim`, ids validated, 0006 retry-safe + checksum-verified + 0001–0005 byte-unchanged). Its **one P2** (a foreign restore materialized data behind a misleading `--reclaim` error) and the actionable **P3s** are fixed here: restore-specific error + manual-`DROP` docs; upgrade-claim docs; a comment on the intentional export pre-barrier placement. Gates green: `test` **4870 passed / 6 skipped**, `quality` + `secret-scan` (tree + history) clean. **Recommend a `/code-review ultra` on this PR** as the final independent check before merge.

## Residual / deferred (host-gated, E06)
- Two installations racing the **very first** claim is a lock-free `INSERT`, not a compare-and-set: `ReplacingMergeTree(claimed_at)` converges to a single owner and the loser is refused fail-closed on its next `require_owner`, but the sub-merge window is real ClickHouse behaviour, not modelled offline. Same-host race is already fenced by the exclusive barrier.
- The one unvalidated interpolated field (`milhouse_version`) is a trusted build-time constant and mirrors the existing `runner.py` ledger insert; bounding it in **both** is a separate follow-up, out of this increment's scope.

## NOT claimed
No change to **G04b** status — the real BACKUP/RESTORE round-trip and the live ownership race stay owner/host-gated (E06, #108). **G04b remains In progress / not passed.**

## Base
Stacked on **#126** (`claude/w04-g04b-restore`); retargets to `main` once #126 merges.

Closes #127
Refs #93, #108

### DIFF
--- a/ops/clickhouse/README.md
+++ b/ops/clickhouse/README.md
@@ -33,6 +33,36 @@ $ milhouse --config ./config.toml storage status    # read-only: applied vs pend
 $ milhouse --config ./config.toml storage migrate    # apply the packaged migrations
 ```
 
+## One ClickHouse per installation (enforced)
+
+Milhouse's single-writer correctness rests on a machine-local commit barrier that fences exactly one
+installation's writers. That barrier cannot see a **second** installation — a different state root,
+clone, or host — pointed at the **same** ClickHouse, so two installations sharing one destination
+would silently comingle and corrupt each other's data. The supported model is therefore **one
+ClickHouse per installation**, and it is **enforced fail-closed**:
+
+- `storage migrate` **claims** the destination for the local installation (a single-owner
+  `_installation` row). A first migrate stamps ownership; a re-run by the same installation verifies.
+- `storage export`, `storage backup`, and `storage restore` **refuse to run** unless the destination
+  is owned by the local installation. A second installation pointed at a claimed destination — or a
+  restore of another installation's backup, whose restored owner row names that other installation —
+  fails closed with `MH_STORAGE_OWNERSHIP` (exit 1) rather than writing or comingling data.
+- An **unclaimed** destination (never migrated) likewise refuses export/backup: run `storage migrate`
+  first.
+- On **upgrade of a pre-existing destination** (one migrated before this guard existed, so it has no
+  `_installation` row yet), the **first** installation to `storage migrate` claims it — ownership is
+  attributed to whoever migrates first, not retroactively to the original writer. In the supported
+  one-per-destination model that is the sole rightful install; a destination that was already
+  (incorrectly) shared is taken back by the rightful owner with `--reclaim`.
+
+To **deliberately re-point** an installation at a new ClickHouse host (or take over a destination that
+was claimed by an installation you are decommissioning), migrate with `--reclaim`, which supersedes
+the prior owner:
+
+```console
+$ milhouse --config ./config.toml storage migrate --reclaim    # take ownership of this destination
+```
+
 ## Back up and restore (native BACKUP/RESTORE)
 
 This deployment provisions a ClickHouse `backups` disk in `config.d/backups.xml` (declared under
@@ -51,6 +81,12 @@ $ milhouse --config ./config.toml storage restore nightly_2026_08_10     # RESTO
   `storage restore` refuses (issuing no drop and no restore) if the analytical database already holds
   a schema. Restore into a clean/fresh state root, or after intentionally dropping the lost database.
   In-place **overwrite with rollback is deferred to W16** — this command never overwrites live data.
+- **A foreign or failed restore leaves the destination populated.** If the restored backup was taken
+  by a *different* installation (its `_installation` owner row names another install), `storage
+  restore` fails closed (`MH_STORAGE_RESTORE`, exit 1) **after** the native `RESTORE` has already
+  materialized that data into the (empty) target. Recovering the clean precondition is a **manual
+  `DROP DATABASE`** — automatic rollback of a failed/foreign restore is deferred to W16. (Restoring
+  your OWN backup verifies its owner row and proceeds.)
 - **Retention and permissions.** The `backups` volume grows with each archive; rotate and remove old
   archives on your own schedule (there is no automatic pruning here). The directory is written by the
   ClickHouse server user inside the container and is loopback-only — never exposed off-host. Treat
--- a/src/milhouse/cli/root.py
+++ b/src/milhouse/cli/root.py
@@ -502,9 +502,14 @@ def storage_plan_command(state: CliState, as_json: bool) -> None:
 
 
 @storage_group.command(name="migrate")
+@click.option(
+    "--reclaim",
+    is_flag=True,
+    help="Take ownership of a destination currently claimed by another installation.",
+)
 @click.option("--json", "as_json", is_flag=True, help="Emit the result as stable JSON.")
 @click.pass_obj
-def storage_migrate_command(state: CliState, as_json: bool) -> None:
+def storage_migrate_command(state: CliState, reclaim: bool, as_json: bool) -> None:
     """Apply every pending migration in order under the exclusive commit barrier; refuse a tamper.
 
     The whole apply runs while THIS installation's control-plane commit lock is held EXCLUSIVELY, so
@@ -516,13 +521,25 @@ def storage_migrate_command(state: CliState, as_json: bool) -> None:
     altered applied checksum is still refused. Requires an initialized control plane (the commit
     lock lives under its control dir).
 
+    After the migrations apply (so the ``_installation`` control table exists), this CLAIMS the
+    destination for THIS installation: "one ClickHouse per installation" is the supported model, and
+    the claim makes the local commit barrier a correct single-writer exclusion. A first migrate
+    stamps the local installation as owner; a re-run by the same installation verifies (a no-op). A
+    destination already owned by a DIFFERENT installation is refused fail-closed (exit 1) UNLESS
+    ``--reclaim`` is given — a deliberate re-point (moving this install to a new ClickHouse host),
+    which supersedes the prior owner. On UPGRADE of a pre-existing destination migrated before this
+    guard existed (no ``_installation`` row yet), the first install to migrate claims it — ownership
+    is attributed to whoever migrates first, not retroactively to the original writer; in the
+    supported one-per-destination model that is the sole rightful install, and a destination already
+    (incorrectly) shared is taken back by the rightful owner with ``--reclaim``.
+
     NOTE: this fence holds only because ``storage export`` is the SOLE production ClickHouse writer
     and it delivers under ``barrier.shared()``. ANY future ClickHouse writer MUST likewise take the
     shared side of this same commit lock, or it could race a migration and lose data.
     """
 
     paths = _resolve_paths(state)
-    _require_installation_id(paths)  # the commit lock lives under an initialized control dir
+    installation_id = _require_installation_id(paths)  # the commit lock lives under a control dir
     database, client = _storage_client(state)
     barrier = GlobalCommitBarrier(
         paths.state_root / bootstrap.CONTROL_DIRNAME / bootstrap.BARRIER_NAME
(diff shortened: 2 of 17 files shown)


that1guy15:
## Independent Milhouse gate review — **FAIL**

- Exact head reviewed: `234e80b27f0309a4e063d8bd2a267c736e395819`
- Reviewed base: `bad5b874e093a4efee0a448c13850d63aab19b15`
- Integrated candidate: the head is one commit whose parent is the reviewed stacked base; all 17 changed files and the inherited W04/G04b recovery path were reviewed.

### Findings

#### [P1] The ownership row is a revocable check, not a destination-scoped fence

Action: Change

Gate/contract: issue #127's fail-closed single-installation invariant; plan §§4.3 and 10.3 writer/maintenance exclusion; G04b recovery integrity.

Evidence: migration 0006 stores competing claims in a `ReplacingMergeTree(claimed_at)` (`src/milhouse/storage/migrations/0006_installation_ownership.sql:16-24`). `claim()` performs a separate `read_owner()` followed by an unconditional insert (`src/milhouse/storage/ownership.py:126-159`), so two first claimants can both observe `None` and both return success. More importantly, `storage export` checks ownership exactly once before entering its state-root-local replay/barrier operation (`src/milhouse/cli/root.py:799-831`). A different installation has a different local barrier and can insert a later `--reclaim` after that check; the first exporter retains no lease, generation, or fencing token and keeps writing. Backup has the same one-shot check inside installation A's local exclusive barrier (`src/milhouse/cli/root.py:620-627`), which installation B does not share.

Two independent exact-head interleavings reproduced the defect offline. Forcing both first-claim reads to observe the empty row produced `FIRST_CLAIM_RACE successful_claim_calls=2`. Separately, installation A passed `require_owner`, installation B reclaimed, and A then wrote a record while the durable owner was B: `STALE_OWNER_WRITE a_authorized=1 owner_before_a_write=installation_B a_record_written=1`. This is the exact check→replay gap in the CLI; the fake was used only to schedule the public operations deterministically.

Impact: during the admitted first-claim window or any `--reclaim`, two installations can both believe they are authorized and overlap export with export/backup/migrate/restore. The new row eventually converges to one displayed owner, but it does not prevent the losing/stale actor from writing, so cross-installation comingling and inconsistent backup/checkpoint evidence remain possible. The PR's statement that the loser is refused on its *next* check does not protect the operation already in flight.

Exact correction: make ownership authority destination-scoped and held for the full operation. Use an atomic claim plus a destination-authoritative lease/fencing generation that every ClickHouse writer and maintenance command validates/enforces through completion, or remove reclaim and use another fail-closed protocol that proves the previous installation is stopped before transfer. Do not treat `ReplacingMergeTree` convergence or a pre-operation read as mutual exclusion. Re-read and verify the authoritative result after any claim, including first claim and reclaim.

Required regression evidence: two independently initialized state roots/clients against one real database, with deterministic races at both first-claim reads, after owner validation, during every exported table insert, around backup snapshot/native statement, and across restore/migration boundaries. Prove the loser blocks or fails before any write and that ownership transfer cannot overlap an old owner's in-flight operation. Exercise clock skew/equal claim timestamps and both supported ClickHouse lines.

#### [P1] The stacked integrated head still false-passes restore reconciliation against replacement spool bytes

Action: Change

Gate/contract: plan §4.3 binds durable segment bytes to the SQLite ledger's header/count/content/file digests; G04b requires restored record IDs and exporter checkpoints to match.

Evidence: PR #128 is stacked on the still-failing PR #126 base and does not change `reconcile_delivered`. At this exact integrated head, `src/milhouse/storage/backup.py:173-184` selects a delivered ledger row but then trusts the caller-root-derived file after only canonical/installation validation. It neither binds `spool_root` to this control database's canonical state root nor compares the parsed file's whole-file digest, byte size, content digest, or frame count to the selected `SegmentRecord`.

An independent exact-head adversarial test committed and delivered record A, replaced `pending/<day>/batch-a.jsonl` with a different canonical same-installation/same-batch segment containing record B, and supplied a restored snapshot containing B but not A. `reconcile_delivered` returned before the assertion failed: `SHADOW_RECONCILE_FALSE_PASS authoritative_missing=1 shadow_present=1 ledger_file_mismatch=1`.

Impact: the production restore command can emit `reconciled=true` while an acknowledged/delivered record is absent. The ownership guard does not repair that inherited broken-recovery path, so the stacked integrated candidate remains unsafe even if the ownership race is corrected.

Exact correction: bind the supplied spool root to the control database's canonical `<state_root>/spool`, then require each trusted parsed segment to equal the authoritative ledger row (whole-file digest, byte size, content digest, frame count, batch/header identity, and applicable exporter state) before using any record ID. Reuse the replay verifier rather than defining a weaker recovery check.

Required regression evidence: wrong-root rejection; canonical same-installation/same-batch replacement bytes; count/content/file-digest mismatches; and a command-level restore showing that a restored set containing only the shadow ID cannot report reconciliation success.

#### [P2] Issue #127 is absent from the contributor-visible Roadmap

Action: Verify

Gate/contract: `AGENTS.md` requires every feature/security issue to be linked into owner Project #1 with `State`, `Kind`, `Work Package`, `Gate`, and `Depends On` fields before work begins.

Evidence: live issue readback for #127 returned `projectItems: []`. The PR closes #127, but no Roadmap item or required field values exist.

Impact: contributor-visible dependency/gate state cannot track this security-sensitive W04/G04 work or make the prescribed merge transition reliably.

Exact correction: add #127 to **Milhouse OSS 1.0 Roadmap** and populate the required fields consistently with W04/G04b and its dependencies, while leaving `docs/implementation-status.md` authoritative.

Required regression evidence: live Project readback showing the issue and all five required field values.

### Test and check evidence

- Exact-head full suite through repository-required `uv 0.11.29`, Python 3.14.6: **4,870 passed, 6 skipped**. Five ClickHouse live tests were skipped because no live server was authorized/available.
- Focused ownership/CLI/backup/migration/packaging/live-definition suite: **87 passed, 5 skipped**.
- Security-critical ownership module coverage: **100% statements and branches** over the unit/CLI suites.
- Independent adversarial reproductions: **three** — two successful first claims from the same empty observation; a stale authorized writer after ownership transfer; and shadow-spool restore reconciliation false-pass.
- Formatting, Ruff, mypy (91 source files), lock, repository/config validation, workflow validation, zizmor, skill validation, private-identifier scan, and tree/history secret scans passed.
- `docs-check`: **72 files / 301 links / 80 external probes** passed; `git diff --check` passed; the exact-head worktree remained clean.
- Independent DCO range check: **1 commit passed**; the PR tip is the inspected commit and its parent is the reviewed base.
- GitHub: **20/20 check runs terminal and successful**, including `required-ci`, test, compatibility 3.11–3.14, package/artifact smokes, gitleaks, audit, dependency review, both CodeQL surfaces, and DCO. Legacy commit statuses: 0.

### Missing or externally pending evidence

- E06 real concurrent ownership/transfer behavior, native backup/restore, merge/TTL fidelity, supported-host clean recovery, and ClickHouse 25.8/26.3 compatibility remain unverified. The checked-in live ownership case is sequential and does not exercise the race that controls acceptance.
- Full backup-time record/checkpoint equality and the composite SQLite/spool manifest remain W16 work, but the bounded W04 reconciliation already shipped must still bind to authoritative bytes and cannot false-pass.
- G04b correctly remains In progress; this review does not change the ledger or accept a gate.

### Residual risk

The immutable migration packaging, existence-guarded reads, owner mismatch errors, restore-specific foreign-backup diagnostic, reference docs, and sequential fake/live test shape are useful pieces. The `milhouse_version` interpolation remains a trusted build-constant assumption, and read-only query commands are not ownership-gated. Neither residual changes the two blocking recovery/concurrency failures above.

**Final verdict: FAIL.** Do not merge this exact head or its failed stacked base. Require a corrective SHA that provides full-operation destination fencing, binds restore reconciliation to the authoritative ledger bytes/root, and registers #127 on the Roadmap; then rerun the adversarial/live evidence and request a fresh exact-head review.

Issue #94, opened 2026-08-03
W05 — Runtime, collector registry, canary, alerting, and vertical slice


that1guy15:
Plugin registry, runtime pipeline (validate→redact→spool→export), bounded HTTP client, site-canary collector, alert rules, spool-only + ClickHouse modes. First end-to-end canary→query slice.

---
*Work package:* `W05` · *Gate:* G05 · *Depends on:* G03 + G04 · *State:* Blocked
*Authoritative source:* [implementation-plan.md](docs/implementation-plan.md) / [implementation-status.md](docs/implementation-status.md)

Issue #132, opened 2026-08-14
W05 spine — deferred /code-review max 131 follow-ups (target grammar, redaction re-bound, strict attribution, double-validate)


that1guy15:
Deferred findings from the max-recall review of **#131** (owner chose "fix High + cheap now"; these four are deferred to tracked follow-ups). Part of **#94**. All are **latent** for the shipped spine (no wired collector/plugin/CLI yet) but should be addressed as increments 2–4 land.

### 1. Config-valid target crashes `TargetDescriptorV1` construction (correctness) — belongs in the config layer
`TargetConfig` (name ≤200 code points/no-control; kind/environment `Identifier`) is **looser** than domain `TargetDescriptorV1` (name `TitleV1`; kind/environment `MachineNameV1`). A config that loads (e.g. `environment="prod_"`, or a >255-UTF-8-byte name) makes `_target_descriptor` raise a `ValidationError` on every run → per-collector isolation reports an opaque `MH_INTERNAL_UNEXPECTED`, so the target-scoped collector silently collects nothing. **Fix:** validate target-descriptor compatibility at **config load** (tighten `TargetConfig`'s grammar to the domain's, or validate the descriptor build at load), so a mismatch is a clear config error, not a silent per-run collector failure. `src/milhouse/runtime/pipeline.py:_target_descriptor` + `config/_models.py:TargetConfig`.

### 2. Redaction output can push a field past its domain byte bound (correctness, latent)
`LayeredRedactor` replaces a substring with a longer pseudonym token, so a within-limits tight-bound field (e.g. `TitleV1` ≤255 bytes) can exceed its bound after redaction → `finalize_record` raises → collector `status='error'`. Latent today (canary emits only `event.message`/`FreeTextV1`'s large bound); reachable for `TitleV1` fields as feedback/alert collectors land. **Fix:** re-bound redaction output (or reserve headroom) so redacting an in-bounds field can't fail finalize. `pipeline.py:_redact_draft`.

### 3. `_build_segment` trusts collector-reported target/privacy_class (correctness/hardening)
[...]

PR #134, opened 2026-08-14
feat(http,collectors): W05 increment 2 — SSRF/TLS-guarded HTTP client + site_canary collector

### LINKED ISSUE #133: W05 increment 2 — bounded SSRF/TLS-guarded HTTP client + site_canary collector
Part of **#94** (W05). Second offline slice; plugs the first real collector into the increment-1 runtime spine (#131). Reuses the spine's registry/pipeline. **Does NOT claim G05/G04 passed** — the real-network canary round-trip is host-gated (E06, #108).

## Scope (offline, Required CI)
- **`src/milhouse/http/client.py`** — a bounded outbound HTTP client (over the declared `httpx` dep) with a **fail-closed SSRF guard** as the security centerpiece:
  - **Method allow-list** (GET only for the canary path); **`follow_redirects=False`** (no redirect-based SSRF bypass); **TLS verification required** (`verify_tls=True`), refuse to disable it.
  - **SSRF**: resolve the target host and **refuse any private / loopback / link-local / CGNAT / reserved / multicast / unspecified address** (incl. the cloud-metadata IP `169.254.169.254` and IPv6 equivalents); block non-http(s) schemes and userinfo-in-URL. Prefer **resolve → validate all resolved IPs → connect to the pinned validated IP** (mitigate DNS-rebinding) — or, if pinning is deferred, validate-then-connect and document the residual rebinding window.
  - **Bounded**: request timeout from config; **cap the response body read** (a canary needs only the status/headers — never read unbounded bytes); bounded retries.
  - Compose the existing `privacy.sanitize_url`/`SanitizedUrl` for URL normalization and keep all errors privacy-safe (no credentialed URL, secret, or path in any error/log). Fixed coded errors (`MH_HTTP_*`).
- **`src/milhouse/collectors/site_canary.py`** — the first first-party collector: reads `SiteCanaryCollector` config (`url`, `expected_statuses`, GET-only, `follow_redirects=False`, `verify_tls=True`, `request_timeout_seconds`), performs one GET via the bounded client, and returns a `CollectorResult` of one `EventDataV1(category="availability", status=…)` `RecordDraftV1`. Uses a **deterministic observation coordinate** (e.g. scheduled canary instant + route/url identity, plan §4.2/§736) for the record id. Registers as a first-party type in the registry (`registry.register("site_canary", …)`), so it rides the spine's pipeline into the spool (spool_only) / export (full).

## Tests (offline — no real network)
- SSRF guard: `http://127.0.0.1/`, `http://169.254.169.254/…`, a private (10./172.16./192.168.) host, a link-local/CGNAT/IPv6-ULA/`::1` host, and a public-name-that-resolves-to-private host are all **refused fail-closed**; a public IP is allowed. GET-only enforced; a redirect is NOT followed; TLS-disable refused; a slow response hits the timeout; an oversized body is capped. Use `httpx.MockTransport` / a stubbed resolver so no socket is opened.
- Collector: healthy (status in `expected_statuses`) vs degraded (not) → correct `EventDataV1` draft; a client error → `CollectorResult(status="failed")` with a fixed code, never a raw URL/secret. Integration: a canary config drives the increment-1 pipeline (spool_only) and commits a redacted `event` segment.

## Deliberately NOT in scope (later)
- Alert-rule engine (increment 3); notification-intents (increment 4); the real-network canary round-trip + live SSRF behavior (E06/#108); the `collect` CLI (W06); the deferred #132 items.

Refs #94, #93, #108

that1guy15:
## What & why
Second offline slice of **W05** (epic #94, issue **#133**): the first real collector (`site_canary`) plugged into the increment-1 runtime spine, plus the bounded outbound HTTP client it needs. The real-network canary round-trip stays owner/host-gated (E06, #108) — **G05/G04 NOT claimed passed**.

## Changes
- **`src/milhouse/http/client.py`** `BoundedHttpClient` — GET-only, `follow_redirects=False`, TLS always on, response body **capped** (streamed), retries only on transport faults. **SSRF guard**: resolve host via an **injectable resolver** → refuse any non-public address (loopback/private/link-local incl. `169.254.169.254`/reserved/multicast/unspecified + CGNAT `100.64/10` + ULA `fc00::/7` + IPv6-embedded IPv4 for mapped/6to4/Teredo/compat) → **pin** the connection to the validated literal IP (no second resolution), preserving the original `Host` + TLS SNI via `sni_hostname` — closing the DNS-rebinding window. Fixed `MH_HTTP_*` errors leak no URL/secret/host/IP/path.
- **`src/milhouse/collectors/site_canary.py`** — the first first-party collector: one bounded GET → one `availability` `EventDataV1` draft (healthy iff status ∈ `expected_statuses`, else degraded), with a deterministic observation coordinate (scheduled instant + sanitized route) for a stable/idempotent record id; a client/SSRF/timeout failure → fail-closed `CollectorResult(status="failed")` with no URL leak. Registers as the `site_canary` first-party type.

## Review & remediation (independent adversarial SSRF review)
Built, then a runtime-grounded (httpx 0.28.1 / httpcore 1.0.9) adversarial SSRF review confirmed the **resolve→validate→pin→SNI core sound** against the full parser/encoding/rebinding matrix (IP encodings, IDNA homographs, embedded-v4 laundering, NAT64 well-known prefix, rebinding — all fail-closed). It drove:
- **P1** — `trust_env=False` on the production `httpx.Client` so `HTTP(S)_PROXY`/`ALL_PROXY` (and `SSLKEYLOGFILE`/`SSL_CERT_*`) can't route a probe through an unvalidated hop and void the pin (+ a `trust_env` test).
- **P2** — reject a non-ASCII (IDN) host with a fixed `MH_HTTP_URL` instead of an uncaught `UnicodeEncodeError`; punycode (`xn--`) hosts accepted (+ tests).
- **P2** — documented the test-only `transport=` injection seam.

Residual (noted, non-blocking): DNS64/NAT64 with a *network-specific* global prefix (environment-specific, requires DNS64+NAT64 in-path, not attacker-general).

## Tests (offline, hermetic — injected resolver + `httpx.MockTransport`, no socket)
Exhaustive SSRF matrix (`tests/security/test_http_ssrf_boundary.py`), client behaviour (pin/Host/SNI, GET-only, no-redirect, TLS-required, timeout, body-cap, retries, IDN reject/punycode accept, `trust_env`), collector healthy/degraded/fail-closed + record-identity determinism, and integration through the runtime spine (spool_only).

## Gates
`test` **5008 passed / 6 skipped** (http + collectors packages **100% line+branch**, critical-coverage-enumerated) · `quality` (ruff, mypy strict, docs-check, zizmor, skill-check) · `secret-scan` (tree + history) — all clean.

## NOT claimed
No change to **G04/G05** — both stay In progress / not passed; the real-network canary + live SSRF/redirect/timeout behaviour is host-gated (#108).

Refs #94, #93, #108
Closes #133


that1guy15:
## Independent Milhouse gate review — FAIL

- Exact head reviewed: `26231b4bdfe37afb2cd2110623d8042739ed1a32`
- Current base reviewed: `5dc4186e13708272ece7ff458039bb4c9c27b738`
- Integrated tree: `c50b32860d8a2ef6df8f0f98589e0b2aabb9edb0`
- Scope: all 13 changed files; W05/G05 contracts; current ledger; ADR 0008; security/privacy, SSRF/TLS, identity/idempotency, timeout, plugin, provenance, packaging, DCO, and hosted-check evidence.

### Findings

#### [P0] Secret-bearing URL paths survive the last redaction boundary and enter the durable record
**Action: Change**

**Gate/contract:** Plan §4.7 requires secrets and URL-sensitive values to be removed or pseudonymized before spool/SQLite/ClickHouse; G05 locks `collect -> validate -> redact -> spool`.

**Evidence:** `src/milhouse/collectors/site_canary.py:73-82` sanitizes the configured URL, but `sanitize_url` intentionally preserves its path; `site_canary.py:127-130` then copies that value into `source.observation.parts.route`. The runtime's last pass (`src/milhouse/runtime/pipeline.py:522-549`) scans only enumerated `data` free-text leaves, not the source observation. On this exact tree I configured `https://example.test/health/supersecretvalue123`, registered `supersecretvalue123` as a known secret, ran the collector, then invoked the exact pipeline redaction/finalization path. The finalized wire still contained the secret: `P0_SECRET_AFTER_PIPELINE_REDACTION True`.

**Impact:** A token or credential embedded in a health URL path is durably written to the spool and can flow to ClickHouse despite the mandatory first-write privacy boundary.

**Exact correction:** Never persist the sanitized URL text as the observation route. Derive a bounded installation-keyed route identifier through a runtime-owned privacy API, and make the runtime traverse or fail closed over every string-bearing draft field before identity/finalization.

**Required regression evidence:** Plant known-secret, credential, encoded, Unicode, host, path, query, fragment, and structured-field canaries in a canary URL/draft and prove absence from finalized records, JSONL spool bytes, SQLite/ClickHouse projections, diagnostics, errors, and logs.

#### [P0] A network-specific NAT64 prefix bypasses the SSRF classifier
**Action: Change**

**Gate/contract:** W05 requires a fail-closed SSRF policy; the issue requires refusal of private/link-local/metadata destinations and their IPv6 equivalents.

**Evidence:** `src/milhouse/http/client.py:99-117` decodes mapped, compatible, 6to4, and Teredo forms only. `client.py:141-157` therefore accepts an otherwise-global IPv6 address under an environment's network-specific NAT64 prefix without checking its embedded IPv4 destination. The exact predicate accepted `2001:4860:64:1::a9fe:a9fe`; on a host whose NAT64 prefix is `2001:4860:64:1::/96`, that translates to `169.254.169.254`. The PR acknowledges network-specific NAT64 as residual, but this is the same metadata SSRF boundary the change claims to close.

**Impact:** On a supported NAT64 environment, attacker-controlled DNS/configuration can reach cloud metadata or another forbidden IPv4 service through an accepted global IPv6 literal.

**Exact correction:** Discover or explicitly configure every active NAT64 prefix and decode/reject translated non-public IPv4 destinations before connect; if that proof is unavailable, fail closed for the affected IPv6 path rather than treating `is_global` as sufficient.

**Required regression evidence:** Cover RFC 6052 network-specific `/32`, `/40`, `/48`, `/56`, `/64`, and `/96` forms mapping loopback, RFC1918, CGNAT, link-local metadata, multicast, reserved, and unspecified IPv4, plus a real supported-host NAT64 smoke.

#### [P1] Classification-policy changes create the same record ID with different content
**Action: Change**

**Gate/contract:** Plan §4.2 requires a repeated ID with the same hash to be a no-op and a semantic parser/configuration change to become an explicit identity coordinate, never an unexplained conflict.

**Evidence:** `src/milhouse/collectors/site_canary.py:81-82,165-169` derives `source_generation_digest` only from implementation version and sanitized URL; it omits `expected_statuses`. `site_canary.py:106-107` uses that omitted policy to produce meaningful content. On this exact tree, the same collector, target, URL, scheduled instant, and HTTP 200 observation with `expected_statuses=[200]` versus `[201]` produced `P1_ID_COLLISION True`, different content hashes, and `healthy` versus `degraded` payloads.

**Impact:** A legitimate canary-policy update can turn a repeat/catch-up probe into an ID/hash conflict instead of an immutable new observation.

**Exact correction:** Bind every normalized input that can change emitted content—including sorted expected statuses and relevant request semantics—into a canonical source-generation/configuration coordinate, without publicly hashing secret URL material.

**Required regression evidence:** For each content-affecting canary setting, prove identical semantics yield identical ID/hash and any semantic change yields a distinct record ID; include same-instant config-change/retry cases.

#### [P1] The request timeout does not bound DNS, the whole response, or retries
**Action: Change**

**Gate/contract:** Plan §4.8 says `request_timeout_seconds` bounds individual source I/O, and W05 requires a bounded shared client and timeout behavior.

**Evidence:** `src/milhouse/http/client.py:283-290` resolves before `_probe`; `client.py:308-316` calls unbounded synchronous `socket.getaddrinfo`; and the monotonic timer starts only at `client.py:340`, after DNS. The `httpx.Timeout` is phase/inactivity based, while `client.py:371-376` has no absolute deadline during body iteration; retries receive a fresh full timeout. Exact-tree fakes showed a 1-second budget succeeding after 10 simulated DNS seconds while reporting `elapsed_ms=0`, and a three-chunk 0.9-second slow drip succeeding after `elapsed_ms=2700`.

**Impact:** A hostile resolver or slow-loris response can monopolize the serial collector pipeline far beyond the configured budget; with retries, the overrun multiplies.

**Exact correction:** Enforce one absolute monotonic deadline across resolution, every attempt, connect/TLS/headers, and body iteration, with bounded/cancellable DNS and remaining-budget propagation. Do not reset the deadline on retry.

**Required regression evidence:** Deterministic DNS stall, sub-timeout slow-drip body, TLS/headers stall, and multi-retry tests must terminate with `MH_HTTP_TIMEOUT` within a documented small scheduling tolerance of the single configured budget.

#### [P1] The delivered first-party collector still does not implement plugin API v1
**Action: Change**

**Gate/contract:** Plan §4.8 and ADR 0008 require every first- and third-party collector to expose `metadata`, `config_model`, `collect`, `health`, `fixture_contract`, and `privacy_manifest`; results also carry cursor, retry classification, and source freshness.

**Evidence:** `src/milhouse/collectors/site_canary.py:58-195` exposes only `descriptor` and `collect`; the runtime protocol in `src/milhouse/runtime/registry.py:41-52` accepts only those two members; and `src/milhouse/runtime/result.py:30-70` has no cursor, retry classification, or freshness fields. Registration therefore declares a W05 first-party collector while bypassing the locked API contract.

**Impact:** The public extension contract cannot be tested or relied upon, and the runtime cannot enforce capability/privacy declarations, health, retry, cursor, or freshness semantics.

**Exact correction:** Implement the complete v1 collector API/result schema and make registry binding validate the exact contract for built-ins and plugins before collection.

**Required regression evidence:** Run the public plugin contract kit against `site_canary` and adversarial incomplete/malformed built-in and entry-point implementations; each missing or inconsistent member must fail closed before collection.

#### [P1] Same-observation retry is still an error, not the required idempotent no-op
**Action: Change**

**Gate/contract:** Plan §4.2 requires repeated ID+hash to be a no-op; the PR claims a stable/idempotent record per scheduled probe.

**Evidence:** `src/milhouse/collectors/site_canary.py:10-13` deliberately generates the same record identity for the same scheduled probe, but `src/milhouse/runtime/pipeline.py:552-559` deliberately derives the same batch name and documents that the retry returns `MH_SPOOL_EXISTS` rather than succeeding as a no-op. This inherited spine path is unchanged by the PR.

**Impact:** Scheduler replay/retry reports a healthy already-committed probe as an error, violating the immutable-record idempotency contract and producing false operational failure.

**Exact correction:** Under the commit barrier, treat an existing segment as success only after verifying exact header, ordered record IDs/hashes, bytes, and delivery obligation; otherwise return a conflict without overwriting.

**Required regression evidence:** Same scheduled probe before/after restart must no-op with one durable record; altered content, config generation, target, exporter set, or bytes under the same candidate ID must fail as a conflict.

#### [P2] Contradictory redirect/TLS configuration validates and is silently ignored
**Action: Change**

**Gate/contract:** The v1 config schema is exhaustive and strict; `site_canary` locks `follow_redirects=false` and TLS verification.

**Evidence:** `src/milhouse/config/_models.py:472-477` types both values as unrestricted `bool`. The exact model accepted `follow_redirects=True, verify_tls=False`, while `site_canary.py:92-97` ignores both values and forces the safe behavior.

**Impact:** `config validate` approves a configuration whose declared behavior differs from runtime behavior, undermining configuration truth and operator diagnostics.

**Exact correction:** Make these locked literals (`Literal[False]` / `Literal[True]`) or reject contradictory values with value-safe validation errors, and ensure the generated schema reflects the fixed semantics.

**Required regression evidence:** Config model, TOML loader, JSON Schema, and CLI validation must reject both contradictory settings without echoing the supplied URL.

### Missing / externally pending evidence and authority

- The current ledger still marks W04/G04 **In progress** and W05 **Pending**, while W05 depends on both G03 and G04. Roadmap epic #94 is still `Todo`; issue #133 has no `type:` label and is not on the Roadmap. This is not a dependency-ready work item under the repository workflow.
- The reviewed base `5dc4186e13708272ece7ff458039bb4c9c27b738` is the tree-identical squash merge of the previously reviewed FAIL spine head; this diff does not remediate the inherited attribution/retention/runtime-contract blockers.
- The real-network SSRF/TLS/redirect/timeout and canary-to-ClickHouse evidence remains host-gated under #108. That pending evidence cannot cure the deterministic failures above.

### Verification evidence

- Exact PR commit tip matched the head; exact head tree `c50b32860d8a2ef6df8f0f98589e0b2aabb9edb0`; protected `main` matched the reviewed base.
- Reviewed every changed file and the current plan, ledger, architecture, ADR 0008, security, provenance, and repository workflow contracts.
- Local focused runtime/HTTP/canary suites: **138 passed**. Full suite: **5,014 collected, exit 0** (consistent with **5,008 passed / 6 skipped**). `quality`, `docs-check`, `skill-check`, `secret-scan`, `git diff --check`, and local DCO all passed.
- GitHub: **20/20** check runs terminal/successful; no legacy commit statuses. Hosted DCO passed. The commit is GitHub-signature `unsigned`, which is not a DCO failure because its Signed-off-by trailer independently passed.
- Adversarial exact-tree results: secret remained after the pipeline redaction/finalization boundary; policy variants collided on record ID with different hashes; DNS and body-drip cases exceeded a one-second budget while succeeding; network-specific NAT64 metadata mapping passed the address predicate; contradictory TLS/redirect config validated.
- Provenance: the public ledger already lists the canary donor path as adaptation-only/not imported; no provenance file changed in this PR. I did not treat that declaration as proof of correctness.

### Residual risks

No live network or external-provider call was made. Real TLS/SNI, redirect, resolver, and NAT64 behavior remains unverified on a supported host, and the W05 alert/notification/full-mode gate is explicitly incomplete. These are additional risks, not reasons to downgrade the reproduced findings.

## Final verdict: FAIL

Do not merge this exact head. Required next action: produce a corrective SHA resolving every P0/P1 above and the inherited spine blockers, add/register #133 (and the existing #132 follow-up) with the required labels/Project fields, and either pass G04 or obtain an explicit plan/ADR dependency amendment before W05 proceeds. Then request a fresh independent exact-head review.

